PR #10, opened 2015-03-12
fix MySQL link name, list links and list rendering

mweirauch:
* suffixing .html to the links was required locally in
  "middleman server" mode


Soulou:
Thank you very much :-)

PR #14, opened 2015-03-13
Detecting HTTPS requests

mweirauch:
* Due to the missing "Internals" section on the front page I put
  that into the "Getting started" section for now.
* Additionally fixed the links to "Getting started" entries where
  the target is not a directory with "index.html" but a concrete
  *.html file which "middleman server" is not able to deliver in
  contrast to github (jekyll?).

### DIFF
--- a/source/index.html.haml
+++ b/source/index.html.haml
@@ -10,11 +10,13 @@
         %li
           %a{href: "/heroku-compatibility"}Heroku compatibility
         %li
-          %a{href: "/internals/procfile"}Procfile - How your app is started
+          %a{href: "/internals/procfile.html"}Procfile - How your app is started
         %li
-          %a{href: "/app/environment"}Environment - Configure your app
+          %a{href: "/internals/detecting-https-requests.html"}Detecting HTTPS requests
         %li
-          %a{href: "/app/domain"}Custom Domain Name
+          %a{href: "/app/environment.html"}Environment - Configure your app
+        %li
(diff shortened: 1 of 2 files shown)


source/index.html.haml
```diff
@@ -10,11 +10,13 @@
         %li
           %a{href: "/heroku-compatibility"}Heroku compatibility
         %li
-          %a{href: "/internals/procfile"}Procfile - How your app is started
+          %a{href: "/internals/procfile.html"}Procfile - How your app is started
```
Soulou: Actually as strange as it is, the link doesn't work in development, but when deployed it's ok, that's why I've let it like this.

Soulou:
I'm going to write really soon an article about how our reverse proxy is working. Your page is directly related to that and it will be perfectly fit it.

Thank you for your work we are greatful for it, I'll work on this later today and will  send a pull request on your branch.

mweirauch:
I assume the github gh-pages service is appending ".html" to the uri in case it does not find a directory (with an index.html in it) with such a name.
My reasoning was that middleman automatically appends ".html" on the generated recent activity links so I thought manually created links should carry the ".html", too.

Soulou:
Fair enough, I'll take care of that too (if I can change the configuration of middleman or the links of the links where this use case happens)

Soulou:
Follow up there: https://github.com/Scalingo/documentation/pull/15

PR #15, opened 2015-03-23
Complete 'detecting HTTPS' from the app

Soulou:


Soulou:
Sorry it took more than 1 day, I've added a go example of code too (Java is not so common finally) And I've merged master into over your commit @mweirauch

mweirauch:
I just wen't through the nightmare again getting middleman up and running on Windows, but that's another story... ;)

Looks good!

Soulou:
So I merge it, thank you very much for you contribution

PR #72, opened 2015-11-30
Fixed anchor were not working

Zyko0:


yannski:
Not really OK with this... :/

Soulou:
It seems to be a limitation of MDL, it's not working either on there website without this kind of nasty hack...

Zyko0:
Is it okay now? There's one more commit not directly linked to the PR

PR #93, opened 2016-03-07
fix: header names

analog-nico:
The header names use lower case only.

### DIFF
--- a/_posts/internals/2015-03-22-routing.md
+++ b/_posts/internals/2015-03-22-routing.md
@@ -12,10 +12,10 @@ containers of your application.
 
 ## Headers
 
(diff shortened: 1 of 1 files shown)


Soulou:
Hi there,

Thanks for your pull request. Your language/framework may downcase everything, but as it is send by our reverse proxy, and as defined in [RFC 7239](https://tools.ietf.org/html/rfc7239) the capital letters are present.

-- Leo

analog-nico:
Oh, I see. I am using node.js with the Express framework by the way. Maybe you can add a note about that...

Soulou:
http://doc.scalingo.com/internals/routing.html#headers done :-)

analog-nico:
👍

PR #105, opened 2016-07-28
Add wordpress deployment

johnsudaar:


_posts/languages/php/2016-07-28-php-wordpress-deployment.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+## Plugins and updates
+
+Since the container filesystem is volatile, plugins and addon should
+be installed and updated within your git repository and never via
+the web interface. You must de-activate autoupdate of all
+your wordpress components.
+
+To do that just add the following lien to your `wp-config.php`
```
yannski: lien -> line

_posts/languages/php/2016-07-28-php-wordpress-deployment.md
```diff
@@ -0,0 +1,81 @@
...
+
+```php
+define( 'AUTOMATIC_UPDATER_DISABLED', true );
+```
+
+## Uploads
+
+The container filesystem is volatile and not synchronised through
+all your instances. So the uploads should not be stored on
+the filesystem itself. We recommend using an external service like amazon s3
```
yannski: can you recommend a plugin to do that?

johnsudaar: There is this one which looks good : https://fr.wordpress.org/plugins/amazon-s3-and-cloudfront/
But i must test it and i do not have the setup to do it here.

PR #116, opened 2016-11-02
Fixed the .extra.paas.php-config example

cristianciofu:
added the missing [  ] from the example


Soulou:
Thanks, that's right, the example was wrong.

PR #119, opened 2016-12-02
PgSQL dump and restore doc match what we do when upgrading

EtienneM:


Soulou:
Can you match the commands in https://github.com/Scalingo/scalingo_databases_rake_tasks-gem/blob/master/lib/scalingo_databases_rake_tasks/tasks/postgresql.rake (with the flag -N to exclude private schema instead of including only the `public` one)

yannski:
LGTM

yannski:
@EtienneM maybe add the "-O" when restoring to avoid user mismatch?

EtienneM:
`-O` is the `--no-owner` that we already use

Soulou:
LGTM

PR #128, opened 2017-03-10
Add review apps documentation

EtienneM:
Related to Scalingo/api#594


yannski:
Thanks for this PR. After reading it, I think there are two different terms that should be clarified: Review App and Child App. Maybe they designate the same thing, maybe not. At least the two terms should be defined at the beginning of the page.

EtienneM:
Merged into #130

Issue #163, opened 2017-08-03
Empty page about dashboard


MrMourse:
https://documentationscalingo-pr154.scalingo.io/articles/dashboard/

EtienneM:
@yannski is it still relevant?

yannski:
Should be rethink. Will open a new ticket if needed.

PR #174, opened 2017-08-08
dashboard page #163

MrMourse:
issue #163


MrMourse:
Don't have the access to cdn.scalingo
dashboard
panel
panel_inf
setting

PR #180, opened 2017-08-09
Adds deployment's documentation page

bragagia:


yannski:
My idea for this page is that it should explain the whold deployment/build process. Therefore, it should mention a few stuff that are missing ATM: postdeploy hook, rolling deployment, TCP SYN connection.

bragagia:
@yannski What are 'rolling deployment' and 'TCP SYN connection' ? :)

PR #185, opened 2017-08-09
Add collapse table of content

magdeleine:
Base on fix/docBootstrap


yannski:
Doc has been redesigned

PR #224, opened 2018-02-09
Rephrase the languages pages

EtienneM:


_includes/nginx_includes.md
```diff
@@ -0,0 +1,20 @@
+This process requires you to edit the `composer.json` file of your project.
+Edit the file the following way:
+
+```json
+{
+  ...
+  "extra": {
+    "paas": {
+      "nginx-includes": ["{{nginx-include}}"]
```
Soulou: Would people understand your `{{nginx-include}}` Is it not better to write an example of path?

PR #279, opened 2018-03-12
reorganized platform section

yannski:


redirections.yml
```diff
@@ -1,4 +1,46 @@
 "301":
+  -
+    old: "/languages/ruby/private-gem-from-git-on-scalingo/"
+    new: "/languages/ruby/private-gem-from-git-on-scalingo"
+  -
+    old: "/platform/app/transfer-ownership"
+    new: "/platform/app/lifecycle-management#transfer-ownership-of-an-application"
+  -
+    old: "/platform/app/application-lifecycle-management"
```
Soulou: Double `old`

yannski: negative

yannski: whoops, you're right

redirections.yml
```diff
@@ -1,4 +1,46 @@
 "301":
+  -
+    old: "/languages/ruby/private-gem-from-git-on-scalingo/"
+    new: "/languages/ruby/private-gem-from-git-on-scalingo"
+  -
+    old: "/platform/app/transfer-ownership"
+    new: "/platform/app/lifecycle-management#transfer-ownership-of-an-application"
```
Soulou: Invalid destination:

https://documentation-service-pr279.scalingo.io/platform/app/lifecycle-management#transfer-ownership-of-an-application

Soulou: Fixed in previous push

Soulou:
good job LGTM

PR #321, opened 2018-03-30
Search revamp

yannski:


yannski:
Serahbox doesn't look great on mobile. Still a bit of work.

Soulou:
Labeled `work in progress` then, but I like it already :-)

yannski:
Now it's better thus reviewable :)

Soulou:
:+1:

PR #332, opened 2018-04-20
Add doc about Express static files

### LINKED ISSUE #328: Deploying an express site to serve a static site
```js
var express = require('express');

var app = express();
app.use(express.static(__dirname + '/dist'));

var port = process.env.PORT || 3000;
app.listen(port, function () {
  console.log('Listening on', port);
});
```

EtienneM:
Fix #328


Soulou:
Add a node about the `dist` directory, it should potentially be different according to the usecase

yannski:
IMHO this article should be in a non-language section as it's a solution to a common problem. Even though today we recommand express, the technical solution may change in the future.

EtienneM:
@yannski You would add it in platofrm/app?

yannski:
Yes, for example.

Le ven. 20 avr. 2018 à 12:45, Étienne M. <notifications@github.com> a
écrit :

PR #335, opened 2018-04-25
Fix typos in swap chart description

cbliard:


Soulou:
Thanks a lot @cbliard 

Merging once validated in the review app

PR #350, opened 2018-05-25
databases should be first class citizens

yannski:


Soulou:
Please don't CLI addon's name right now, keep `scalingo-*`

PR #353, opened 2018-06-01
Doc custom Node.js cache folder

### LINKED ISSUE #227: Configuration of custom cache directories for node application should be documented

EtienneM:
Fix #227 

https://documentation-service-pr353.scalingo.io/languages/nodejs/start#custom-cache-folder

@Soulou I am not sure this is what you meant for this issue. If it is, what is the purpose of customizing this folder?

### DIFF
--- a/_posts/languages/nodejs/2000-01-01-start.md
+++ b/_posts/languages/nodejs/2000-01-01-start.md
@@ -153,6 +153,21 @@ You may want to run custom commands or scripts before or after the installation 
 }
 ```
 
+### Custom Cache Folder
+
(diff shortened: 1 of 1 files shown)


Soulou:
No it's not this cache I'm talking about, it's the cache directory used for the deployment cache, look at buildpack and look for the cacheDirectories key read in the package.json, by default it puts node_modules and potentially bower_components if Bower is detected, but the idea is to be allowed to add any directory to the deployment cache.

EtienneM:
OK thanks for the clarification. You might find this one more suitable ;)

PR #368, opened 2018-06-15
More info about golang-migrate

### LINKED ISSUE #223: Provide a sample of use of `additional-tools` section in dep page
In [this page](https://doc.scalingo.com/languages/go-dependencies-with-dep), we talk about the `additional-tools` [migrate](https://github.com/mattes/migrate) but I don't really see what it is, And why it's here. We should provide an example to make it clearer

EtienneM:
We should fix Scalingo/go-buildpack#3 before merging this.

There is no tutorial on how to use this lib. But there is a plan to add one: golang-migrate/migrate#40. Then we can write our own tuto

Fix #223


Soulou:
Heroku is doing something right now, let's follow https://github.com/heroku/heroku-buildpack-go/issues/256

PR #370, opened 2018-06-15
Introduce logs drain as a beta feature

### LINKED ISSUE #356: Add page about the log drain feature
With a warning at the beginning stating it's a beta feature

EtienneM:
https://documentation-service-pr370.scalingo.io/platform/app/log-drain

@Soulou Do you want to enforce an environment variable name for the token, to prepare the release of the non-beta feature? Or it will not be about an environment variable name?

Fix #356

### DIFF
--- a/_posts/platform/app/2000-01-01-log-drain.md
+++ b/_posts/platform/app/2000-01-01-log-drain.md
@@ -18,6 +18,8 @@ work when trying to find an information amongst all these lines of logs. A few S
 such as Logentries, Logmatic.io, Papertrail or Graylog. You can also host such solution on Scalingo.
 Here is a tutorial on [how to deploy the ELK
 stack](https://scalingo.com/articles/2018/02/23/running-the-elk-stack-on-scalingo.html) on Scalingo.
+We also support sending your applications logs using the standard syslog protocol over UDP, TCP or
+TCP+TLS.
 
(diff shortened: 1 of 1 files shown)


Soulou:
For me it should not be in the environment, too clumsy. It will be a custom integration where the user put the URL/token or as an Addon. (I'm in favor of the first)

Soulou:
There is nothing about syslog, it should be present (UDP/TCP/TCP+TLS)

_posts/platform/app/2000-01-01-log-drain.md
```diff
@@ -0,0 +1,27 @@
+---
+layout: page
+title: Add a Log Drain to your App
+nav: Log Drain
+modified_at: 2018-06-15 10:00:00
+tags: Papertrail Logentries Logmatic Graylog logs integration
```
yannski: tags are not used ATM. Theses names should appear explicitely in the "real" content of the page.

EtienneM:
> There is nothing about syslog, it should be present (UDP/TCP/TCP+TLS)

I don't really get what you mean here

Soulou:
We don't support only private providers to drain logs to, but also standard protocol : syslog.

Syslog over yep/TCP/TCP+TLS

PR #371, opened 2018-06-15
Using redis as cache engine for Rails

### LINKED ISSUE #357: Doc about flushing Redis in postdeploy hook
https://app.intercom.io/a/apps/w4oogu7s/inbox/inbox/all/conversations/14845559720

EtienneM:
https://documentation-service-pr371.scalingo.io/languages/ruby/rails/redis-cache

Fix #357

### DIFF
--- a/_posts/databases/redis/2000-01-01-data-import.md
+++ b/_posts/databases/redis/2000-01-01-data-import.md
@@ -6,7 +6,7 @@ tags: databases redis dump restore migration
 index: 2
 ---
 
-This tutorial aims at transfering all the data from a remote Redis database
+This tutorial aims at transferring all the data from a remote Redis database
 (from another provider) to a Redis instance provisioned through the [Scalingo
 Redis Addon]({% post_url databases/redis/2000-01-01-start %}).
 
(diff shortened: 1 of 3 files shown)


Soulou:
I disagree with the title here, it doesn't match what you're explaining.
You're not flushing anything, you're "Using redis as cache engine for Rails"

Soulou:
The directory seems wrong to me now: posts/addons/scalingo-redis/2000-01-01-flush.md

Soulou:
I still don't like that the page is in the addon section while it's a really rails specific issue

Soulou:
@EtienneM what are we doing with this article?

EtienneM:
I don't really know.. The trick is kind of nice to know and it's not only about Rails. We should add example with more languages. The idea is just to show how we could deprecate old cache key in Redis by using some env variable.

Maybe add a link to the "Redis Cache Mode" section makes this part more integrated to the whole doc?

yannski:
Maybe it would be bettre to have an "Engineering blog" for this kind of Tips'n'Tricks?

PR #396, opened 2018-07-26
Add containers limits

### LINKED ISSUE #378: Explain what is CPU consumption, how we share the CPU

### LINKED ISSUE #382: Add a page to document each container size limits
- CPU Access (Related to #378 )
- RAM
- SWAP
- PID limits
- ...

EtienneM:
https://documentation-service-pr396.scalingo.io/platform/internals/container-sizes

A couple of questions for @Soulou:

- What is the difference PIDs limit and ulimit nproc?
- Should we mention the bandwidth limit?

Fix #382
Fix #378

### DIFF
--- a/_posts/platform/internals/2000-01-01-container-sizes.md
+++ b/_posts/platform/internals/2000-01-01-container-sizes.md
@@ -5,7 +5,7 @@ tags: internals containers sizes
 index: 2
 ---
 
-## Comparative table
+## Comparative Table
 
 <table class="table">
 	<thead>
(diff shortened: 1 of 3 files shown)


Soulou:
Only the PID limit is in action right now, so don't speak of the rest, the bandwidth limit is not systematic either yet.

EtienneM:
Actually I read the code to know about the ulimit nproc limit so it seems to be in action

Soulou:
But it's not I promise, it's only some preliminary work

PR #480, opened 2019-01-11
Add mysql 5.7.24 and 5.6.42

johnsudaar:


johnsudaar:
https://doc.scalingo.com/changelog patch are not here

PR #496, opened 2019-02-08
How to use Redis for PHP session

### LINKED ISSUE #476: how to handle PHP session with Redis
https://app.intercom.io/a/apps/w4oogu7s/inbox/inbox/conversation/20136452283#part_id=comment-20136452283-2447789427

EtienneM:
https://documentation-service-pr496.scalingo.io/languages/php/php-session-redis

Fix #476

### DIFF
--- a/_includes/nginx_includes.md
+++ b/_includes/nginx_includes.md
@@ -16,5 +16,7 @@ If you are not using composer, create a `composer.json` file with the previous
 content, and also create a file `composer.lock` containing an empty JSON
 dictionary `{}`.
 
-> Tip: You can find more information about extra configuration in [the PHP
+{% note %}
+Tip: You can find more information about extra configuration in [the PHP
 support page]({% post_url languages/php/2000-01-01-start %}).
+{% endnote %}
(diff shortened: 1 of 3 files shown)


yannski:
Can the 'tcp://host:port?auth=password' string be replaced with the environment variable?

yannski:
I mean, if the goal is to enforce good practices, we can go deeper :)

Soulou:
@yannski I think we could automate this with the buildpack at some point, and it would be much easier for the user, question is about the implementation:

* If a redis addon is present, automatically set the session there?
* If a env variable `PHP_REDIS_SESSION=true` is set and addon is present?
* other?

EtienneM:
> Can the 'tcp://host:port?auth=password' string be replaced with the environment variable?

Not possible to have an env variable in composer.json.

> I think we could automate this with the buildpack at some point

Good idea. I wouldn't go with the first idea. But something like the second one seems like a good idea. Or by specifying it in the composer.json instead of an env variable? Like `.extra.paas.redis_session = true`

EtienneM:
We should probably merge this PR and open a new issue in the relevant repository, don't you think?

yannski:
Done is better than perfect

PR #502, opened 2019-02-22
Remove duplicated changelog

EtienneM:
This page is a duplicate of the changelog.scalingo.com page. We will forget to
update it. For instance, the 1.9.0 has not been announced here.


Soulou:
You can't delete a page without making the proper redirection, can you add it in the `redirections.yml`?

PR #515, opened 2019-02-22
Fix padding

### LINKED ISSUE #462: [Changelog][CSS] Links have a 12px left padding
screenshot from 2018-11-26 09-53-02

It does not really look nice..

johnsudaar:
Fix #462


johnsudaar:
In fact you're right it was used in the side-bar. I checked the entire doc and i did not see any other side effect.

PR #553, opened 2019-04-12
Changed addons' plan names

jonathanfallon:
Addon plans have changed.
I've added the reference to the CLI commands to list the current plans.


Soulou:
Hello @jonathanfallon, you're completely right. Actually there were never a mongo plan named just 'sandbox', it was an error from our side. Our apologies for this and thanks for the contribution

PR #562, opened 2019-05-03
Document NPM_NO_BUILD for Node.js apps

### LINKED ISSUE #546: [Node.js] Document the NPM_NO_BUILD env variable
https://app.intercom.io/a/apps/w4oogu7s/inbox/inbox/conversation/21534245059#part_id=comment-21534245059-2897929693

EtienneM:
Fix #546

https://documentation-service-pr562.scalingo.io/languages/nodejs/start#nodejs-build-hooks

### DIFF
--- a/_posts/languages/nodejs/2000-01-01-start.md
+++ b/_posts/languages/nodejs/2000-01-01-start.md
@@ -141,20 +141,30 @@ Example of `package.json`:
 
 ### Node.js Build Hooks
 
-You may want to run custom commands or scripts before or after the installation of dependencies. This is possible thanks to commands named `prebuild` and `build`. The first is run before installing dependencies and the latter after. To use them, add the commands you wish to run in the `package.json` the following way:
+You may want to run custom commands or scripts before or after the installation
+of dependencies. This is possible thanks to commands named `prebuild` and
+`postbuild`. The first is run before installing dependencies and the latter after.
+To use them, add the commands you wish to run in the `package.json` the
+following way:
 
 ```json
 {
   ...
   "scripts": {
     "prebuild": "node prebuild.js",
-    "build": "grunt build"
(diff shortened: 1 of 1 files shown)


_posts/languages/nodejs/2000-01-01-start.md
```diff
@@ -141,20 +141,30 @@ Example of `package.json`:
 
 ### Node.js Build Hooks
 
-You may want to run custom commands or scripts before or after the installation of dependencies. This is possible thanks to commands named `prebuild` and `build`. The first is run before installing dependencies and the latter after. To use them, add the commands you wish to run in the `package.json` the following way:
+You may want to run custom commands or scripts before or after the installation
+of dependencies. This is possible thanks to commands named `prebuild` and
```
Soulou: It's actually not clear what is the difference between `build` and `postbuild` etc.

Actually there is none, and it's why the previous version of the doc, said `postbuild` are aliases of `build`, it was for the retrocompatibility at some point.

EtienneM: There is at least one difference: you can skip the build phase with the `NPM_NO_BUILD` env variable whereas you can't skip the postbuild phase. So postbuild can't be called an alias of the build phase.

EtienneM: And the comments in the buildpack states for the build phase:

```
    # Run build by default as it is the standard task to
    # bundle applications in most javascript apps.
```

And for the postbuild phase:

```

  # Run postbuild script to trigger tasks unrelated to
  # the build, like notifying some external services or
  # applying migrations

```

Soulou: Those paragraphs should be present in the documentation then

PR #581, opened 2019-05-29
Update 2000-01-01-domain.md

jloosfelt:
First time I tried to setup a wildcard domain, I was unsure of how to get the right code for my domain.


johnsudaar:
Thanks a lot!

PR #584, opened 2019-06-05
Fix typo line 80: missing plural for 'requirement'

laurentP13:
replace 'pip freeze > requirement.txt'
with
'pip freeze > requirements.txt'


yannski:
Thanks!

PR #621, opened 2019-07-10
Use a valid CIDR format

juherr:
Some IT support don't know .0 is not a significative information


EtienneM:
Thanks for the patch

PR #657, opened 2019-08-23
Add doc about new SCM integrations

### LINKED ISSUE #651: Add documentation about new SCM integrations
GitLab, GitLab self hosted, GitHub Enterprise

EtienneM:
To be merged when the feature is in prod

Fix #651


_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
```diff
@@ -158,6 +161,8 @@ You know how to deploy a Ruby app, now develop your project and build something
 
 Here are some links that you could find useful:
 
-* Need to store data? Look at our [database addons](https://scalingo.com/fr/databases)
+* Need to store data? Look at our [database
+  addons](https://scalingo.com/fr/databases)
```
yannski:
```suggestion
  addons](https://scalingo.com/databases)
```

yannski:
@EtienneM when doc page are deleted, a redirection should be added. Take a look at https://github.com/Scalingo/documentation/blob/master/redirections.yml.

yannski:
BTW, for SEO purpose, I'm wondering if separate pages should be added "Github and GitHub Enterprise integration" and "GitLab integration". What do you think?

EtienneM:
> BTW, for SEO purpose, I'm wondering if separate pages should be added "Github and GitHub Enterprise integration" and "GitLab integration". What do you think?

I have no knowledge on SEO. But aren't we punished if there is some duplicate content? And we'll have a lot of duplicate content if we have two pages for GitHub and GitLab.

yannski:
I may be mistaken but the page will be quite different (even tough some wording will be the same)

PR #678, opened 2019-09-13
specific PG databases access page

yannski:


yannski:
IMHO "Access Database" in a generic manner does not really help end users. They look for how to access *their* database not *any database*. That's why I think a dedicated page is more useful for them. My angle is not about the rightness of the informations but how end user look for the information and how they will read the information. What do you think about that?

yannski:
Maybe the page you're linking should be more generic, less database specific details and link to each database page?

EtienneM:
> IMHO "Access Database" in a generic manner does not really help end users. They look for how to access _their_ database not _any database_. That's why I think a dedicated page is more useful for them. My angle is not about the rightness of the informations but how end user look for the information and how they will read the information. What do you think about that?

I agree with you, that's why I would like this PR to add a `Access Your Database` entry to each DB type. But the content is the content we currently have on https://documentation-service-pr678.scalingo.io/platform/databases/access

Soulou:
4 months no news closing

yannski:
I still plan to work on it

ksol:
Should we close this one ?

EtienneM:
I think so

yannski:
Next time please wait for the author's comment; And because of the conversation in this thread I opened this issue: https://github.com/Scalingo/documentation/issues/1314

PR #727, opened 2019-11-15
Add a word about how to set the application timezone

### LINKED ISSUE #718: Documentation to change timezone in the environment of a container should be present
With the environment variable `TZ`

ie. `TZ=Europe/Paris`

EtienneM:
Fix #718


Soulou:
LGTM

PR #730, opened 2019-11-15
[Review App] Protect access with basic auth

### LINKED ISSUE #711: [Review app] Protect with basic auth
https://app.intercom.io/a/apps/w4oogu7s/inbox/inbox/all/conversations/24224179935

EtienneM:
Fix #711

### DIFF
--- a/_posts/platform/app/2000-01-01-review-apps.md
+++ b/_posts/platform/app/2000-01-01-review-apps.md
@@ -85,6 +85,24 @@ Then you can test the presence of this environment variable in your application
 and setup basic auth in case this environment is defined in the environment of
 the application.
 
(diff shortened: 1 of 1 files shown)


Soulou:
Don't you want to explain how to set the password also, this seems a bit incomplete also.

EtienneM:
> this seems a bit incomplete also.

I added a word about the password. Is it what you meant or do you think I should add more? The thing is that the part about setting up basic auth is really dependent of the technology used

Soulou:
I would add something like this for the password, "The environment variable should be used by your app to configure the HTTP Basic auth"

yannski:
LGTM

PR #784, opened 2019-12-23
[changelog] Go 1.13.5 and 1.12.14

EtienneM:
Tweet:
> [Changelog] Buildpacks - Go - New Versions: 1.13.5 and 1.12.14. 1.12.14 is the
> new default. https://changelog.scalingo.com #golang #changelog #PaaS #upgrade


EtienneM:
It's actually already the case since a couple of days so I auto-merge this.

PR #789, opened 2019-12-24
[changelog] PHP 7.4.1, 7.3.13 and 7.2.26

EtienneM:
Tweet:

> [Changelog] Buildpacks - PHP - Support of PHP 7.4.1, 7.3.13 and 7.2.26 https://changelog.scalingo.com #php #changelog #PaaS

Fix https://github.com/Scalingo/php-buildpack/issues/114


EtienneM:
Same than the other, this is already deployed in prod, let's announce it

PR #795, opened 2020-01-10
make autoscaling available upon request instead of private beta 

### LINKED ISSUE #794: Autoscaling is not in beta anymore, it's available on support request

yannski:
fix #794

Motivation: as seen by the emails sent by some customers asking to get access to the feature, the "beta" tag is too deceptive. It implies a lack of stability which is not the case here.

### DIFF
--- a/_includes/upon_request.md
+++ b/_includes/upon_request.md
@@ -1,0 +1,3 @@
+{% warning %}
+  This feature is currently available upon request. Ask the support to get access to it.
+{% endwarning %}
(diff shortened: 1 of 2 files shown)


johnsudaar:
Why ??? Couldn't we just keep it as beta if it's not finished or release it if it's finished ? Why do we need another state in between ?

yannski:
The state hasn't changed. It's handling hasn't changed. In reality, it's not in beta anymore for a loooong time now. @johnsudaar BTW this issue tracker is public.

Soulou:
> Why ??? Couldn't we just keep it as beta if it's not finished or release it if it's finished ? Why do we need another state in between ?

I agree with @yannski here, it's not a beta for a really long time

PR #856, opened 2020-03-05
Specifying the PHP version is done via .require.php

EtienneM:
`.config.platform.php` is not meant for production usage:

https://getcomposer.org/doc/06-config.md#platform

> Lets you fake platform packages (PHP and extensions) so that you can emulate a production env


Soulou:
Heroku seems to prefer `require.php`, shouldn't we do the same?

EtienneM:
> Heroku seems to prefer `require.php`, shouldn't we do the same?

That's exactly what I'm doing in this PR..

Issue #878, opened 2020-03-22
Implement Basic Auth on Review Apps


yannski:
1. It leads to duplicate content which is bad for SEO
2. Some customers may be confused by reading documentation from review apps which may not be up-to-date

Soulou:
It seems that some customers shared review app links in a support session, it's not normal they found it

Soulou:
Probably it comes from the fact that the `documentation` repo is public on GitHub, so Google indexes it, including the commit status which contain the URLs of the review apps, leading to this situation.

PR #883, opened 2020-03-24
Optionally protect by basic-auth

ksol:
Fix for #878


yannski:
Can you add the necessary manifest to activate it by default on Review Apps?

ksol:
It's coming. If the PR isn't mark as `reviewable` it means it's still a work in progress

yannski:
Tips: usually I set the user/password as env variable in the parent (as a single env var) and then declare a second env var that acts as an "activator" in the manifest that will be read by Review Apps

ksol:
Took me a little while to realize that quotes should not be included in the env string...

Anyway, the pass is a:z for this review_app; I'll let you set `BASIC_AUTH` to a suitable duo for the parent app.

ksol:
I suppose this should also be applied to the dev-docs ?

yannski:
@ksol yes, it would be nice to have in on the dev docs as well

PR #884, opened 2020-03-24
Add region migration page

EtienneM:
I decided to not include the complete output of the CLI to shorten the page.

https://documentation-service-pr884.osc-fr1.scalingo.io/platform/app/regions-migration


yannski:
I propose to merge it now. I will propose changes when I will actually use it :)

PR #894, opened 2020-03-30
Fix adminer SNC URL

mackwic:


Soulou:
I'm actually disagreeing, it's a service officially supported by Scalingo, so `.com`

PR #917, opened 2020-04-14
Add Erlang sample

EtienneM:
Waiting for Soulou to accept my ownership transfer request of
https://github.com/EtienneM/sample-erlang so that it gets renamed to
https://github.com/Scalingo/sample-erlang.


Soulou:
Done https://github.com/Scalingo/sample-erlang

PR #950, opened 2020-05-06
mention storky 

yannski:
…g article

### DIFF
--- a/_posts/platform/app/2000-01-01-regions-migration.md
+++ b/_posts/platform/app/2000-01-01-regions-migration.md
@@ -6,6 +6,8 @@ tags: app migration region
 index: 17
 ---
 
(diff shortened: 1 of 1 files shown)


_posts/platform/app/2000-01-01-regions-migration.md
```diff
@@ -6,6 +6,8 @@ tags: app migration region
 index: 17
 ---
 
+Aka *“Project Storky”*.
```
EtienneM, line 9: Not sure it properly displays like that

yannski: You can check here: https://documentation-service-pr950.osc-fr1.scalingo.io/platform/app/regions-migration
Seems good to me, no?

PR #992, opened 2020-06-16
Search result leads to review app and not always prod

EtienneM:
- Works in dev :heavy_check_mark: 
- Works in the review app (https://documentation-service-pr992.osc-fr1.scalingo.io/) :heavy_check_mark: 

Related to https://github.com/Scalingo/one-stop-shop/issues/105


app/javascript/js/search.js
```diff
@@ -21,6 +21,36 @@ if (searchNode !== null) {
...
+      if (context.selectionMethod !== 'click') {
+        console.log("NOT A  CLICK")
+        return
+      }
+
+      // Replace the domain name doc.scalingo.com with the domain name of the
+      // current review app.
+      var protocol;
+      var hostport;
+      if (process.env.JEKYLL_ENV == 'development') {
```
ksol, line 42: That works? 🤔

EtienneM: Héhé no, that's the problem I need to solve now. search.js is executed in the browser. I need to find a way to have access to the environment variable in this script but I'm not sure it's possible :'( Do you have an idea?

ksol: No, not really; you'd have to pass it from rails (that has access to the env) to the js (via a data-attribute somewhere on the dom) or something like that.

In this case, I'd have gone with just using whatever's the current url of the browser (using `window.location` and its properties (I don't know if you tried it, considered it or anything)

ksol: (ok this is not a rails-app, forgot about that; I'm assuming there's some hooks we could use from Jekyll to get it, though)

EtienneM: You're right, `window.location` is probably the way to go. Thanks for the suggestion!

EtienneM:
Closing the PR to re-trigger the deployment of the PR

PR #995, opened 2020-06-16
Add a 'Communication With TLS' section

### LINKED ISSUE #946: TLS paragraph in common Database page
Encryption at Rest is mentionned here:

https://doc.scalingo.com/platform/databases/intro

This page should also include a paragraph about TLS/Force TLS : our vision and the status for each database.

This page should also mention the Internet Accessibility feature and the default behavior.

EtienneM:
Fix #946

### DIFF
--- a/_posts/databases/elasticsearch/2000-01-01-start.md
+++ b/_posts/databases/elasticsearch/2000-01-01-start.md
@@ -122,7 +122,7 @@ $ scalingo --app my-app run curl $SCALINGO_ELASTICSEARCH_URL
 }
 ```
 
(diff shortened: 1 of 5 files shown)


EtienneM:
> Can you elaborate on the specific Elasticsearch case?

What is specific in this case?

yannski:
> > Can you elaborate on the specific Elasticsearch case?
> 
> What is specific in this case?

For ES, it's either TLS or no TLS, you can't have both at the same time. Furthermore, TLS is not available by default. Whereas all other databases always have TLS by default.

PR #997, opened 2020-06-17
[Search] Always update the window.location even in prod

EtienneM:
#992 introduced a bug described in (https://github.com/Scalingo/one-stop-shop/issues/145). What I changed is that we actully need to update the `window.location`, even in prod. Otherwise nothing happens when clicking on a search suggestion.

Related to https://github.com/Scalingo/one-stop-shop/issues/145

### DIFF
--- a/app/javascript/js/search.js
+++ b/app/javascript/js/search.js
@@ -22,5 +22,21 @@ if (searchNode !== null) {
       hint: false,
       openOnFocus: true,
     },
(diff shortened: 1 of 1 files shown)


ksol:
LGTM. We'll have to deploy to prod anyway, to see if it works there

EtienneM:
FYI for the documentation, the `master` branch is the production :)

PR #1010, opened 2020-07-27
Add a note to inform that procfiles do not accepte multiple same name rules

cyrille-sg:
Add a block of text to prevent two targets in a procfile
Code followup: https://github.com/Scalingo/api/issues/1747

### DIFF
--- a/_posts/platform/app/2000-01-01-procfile.md
+++ b/_posts/platform/app/2000-01-01-procfile.md
@@ -35,13 +35,14 @@ The standard syntax is:
 
 ```yaml
 <container type>: <command>
-
-# where container type should contain letters and digits only
-# and command can be any command line which can includes environment variables
 ```
-As a note,`container type` are unique in the `Procfile`. So, you cannot have, for example, two `web` `container type`.
+where container type should contain letters and digits only and command can be any command line which can includes environment variables.
+
+{% note %}
+`container type` are unique in the `Procfile`. So, you cannot have, for example, two `web` `container type`.
 If you have, only the second one will be executed, as if the first did not exist.
 If you want run a command, you may want to look at [Post deploy](https://doc.scalingo.com/platform/app/postdeploy-hook).
+{% endnote %}
 
 When you have more complex commands to run, don't hesitate to write a short start script,
 for instance in `bin/start.sh`:
--- a/_posts/platform/app/2000-01-01-procfile.md
+++ b/_posts/platform/app/2000-01-01-procfile.md
@@ -36,9 +36,12 @@ The standard syntax is:
 ```yaml
 <container type>: <command>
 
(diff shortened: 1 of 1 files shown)


_posts/platform/app/2000-01-01-procfile.md
```diff
@@ -36,9 +36,12 @@ The standard syntax is:
 ```yaml
 <container type>: <command>
 
-container type: should contain letters and digits only
-command: any command line which can includes environment variables
+# where container type should contain letters and digits only
+# and command can be any command line which can includes environment variables
 ```
+As a note,`container type` are unique in the `Procfile`. So, you cannot have, for example, two `web` `container type`.
+If you have, only the second one will be executed, as if the first did not exist.
```
EtienneM, line 43: I wouldn't explain the behaviour in case there is two container type described in the Procfile. We actually don't want anybody to have the same name for multiple container types. Hence we should just say: "don't do that or you may have unexpected behaviour". Don't you think?

cyrille-sg: There is a ticket to add a fatal error in the procfile has two targets with the same name, but it is done yet. I will edit the docs to include the behavior once the ticket is done

EtienneM: Can you add a link to this ticket in this PR so that it is reflected in GitHub UI?

EtienneM: You talk about an issue but I don't see it mentioned in the PR. Can you add a reference to this issue?

EtienneM: You didn't addressed my previous message in this thread.

cyrille-sg: https://github.com/Scalingo/api/issues/1747

cyrille-sg:
https://github.com/Scalingo/api/issues/1747

_posts/platform/app/2000-01-01-procfile.md
```diff
@@ -35,10 +35,14 @@ The standard syntax is:
...
-
-container type: should contain letters and digits only
-command: any command line which can includes environment variables
 ```
+where container type should contain letters and digits only and command can be any command line which can includes environment variables.
+
+{% note %}
+`container type` are unique in the `Procfile`. So, you cannot have, for example, two `web` `container type`.
+If you have, only the second one will be executed, as if the first did not exist.
+If you want run a command, you may want to look at [Post deploy](https://doc.scalingo.com/platform/app/postdeploy-hook).
```
EtienneM, line 44:
```suggestion
If you want to run a command, you may want to look at [Post deploy](https://doc.scalingo.com/platform/app/postdeploy-hook).
```

PR #1020, opened 2020-08-03
Remove the PhantomJS page

### LINKED ISSUE #972: [Meteor] Reference about PhantomJS should disappear, the technology is deprecated
Deprecated since 2018 https://github.com/ariya/phantomjs/issues/15344

EtienneM:
I don't know if the purpose of the related issue is to completely remove the page or replace with an alternative?

Fix #972


Soulou:
The alternative has an issue now: puppeteer

PR #1078, opened 2020-09-30
PHP: bump available version and add intl

Soulou:


EtienneM (changes requested):
Can you also add a changelog entry, and tweet about this?

EtienneM (approved):

PR #1093, opened 2020-10-26
Add internal link from the generic restore backup page

### LINKED ISSUE #890: [Databases] Backup Restauration only provides instruction for MongoDB
We would need to add MySQL and PostgreSQL to the operations.

Source page: https://doc.scalingo.com/platform/databases/restore-backup

EtienneM:
https://documentation-service-pr1093.osc-fr1.scalingo.io/platform/databases/restore-backup#top-of-page

Fix #890

### DIFF
--- a/_posts/platform/databases/2000-01-01-restore-backup.md
+++ b/_posts/platform/databases/2000-01-01-restore-backup.md
@@ -1,6 +1,6 @@
 ---
 title: Restore your Database Backup
-modified_at: 2020-02-13 00:00:00
+modified_at: 2020-10-26 00:00:00
 tags: databases backups
 ---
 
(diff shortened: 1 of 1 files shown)


mackwic:
For a lector skimming over the page, this could be confusing... What about removing the exemple altogether, only keeping the links to specific dbs ? (I'm not sure on that...)

EtienneM:
Here is how both solutions look like. I tend to prefer the solution with no instructions on this page but just link to each DB type page. But I'm open to suggestion.
Screenshot_2020-10-27 Restore your Database Backup
Screenshot_2020-10-27 Restore your Database Backup(1)

mackwic:
Agreed, it's better with only the direct links. Let's do that ! 👍

PR #1118, opened 2020-12-03
Add a doc page about SonarQube on Scalingo

EtienneM:


yannski:
If I may: this doc page does not explain how to install sonarqube in an existing app, which I assume will be the generic use case.

EtienneM:
Do you mean, installing SonarQube in the same container as the production application?

yannski:
> @yannski Do you mean, installing SonarQube in the same container as the production application?

Yes. Maybe it's not the role of this tool?

EtienneM:
Yes it's not intended to run alongside the application but as a separate application. SonarQube is a heavy analysis tool which needs 4 GiB RAM to run. Hence we don't want it alongside the user application.

mackwic:
Shouldn't we advise to fork the repository in order to activate the Github linking ?

EtienneM:
> Shouldn't we advise to fork the repository in order to activate the Github linking ?

I'm not sure I understand what you mean

mackwic:
If I want to have a personal SonarQube, forking the buildpack then deploying the fork makes me able to use the “Manual deploy from Github” feature when Sonar updates, isn’t it ?

It was my understanding of how the Metabase build pack was working….

--
Thomas WICKHAM
Business Development Executive @Scalingo
Website: https://scalingo.com
Telephone: +33 6 30 06 23 15
The automagic cloud for high-performing teams - Le cloud automagique pour les équipes performantes

EtienneM:
This is not how it is working. You can't link a Scalingo application to a buildpack and deploy the application automatically when a new version of the buildpack is pushed. 

You can link a Scalingo application to a repository hosting the code of the application. But in this case, the code of the application is empty (https://github.com/Scalingo/scalingo-sonarqube) except for some configuration files (`.buildpacks`, `scalingo.json`, etc). The [`scalingo-sonarqube`](https://github.com/Scalingo/scalingo-sonarqube) repository is the one you could link your Scalingo application to, but I'm not sure it would be interesting.

mackwic:
It's not only interesting, it’s needed to update the application. The ability to update the application is not a nice-to-have, it’s mandatory.

So from what you’re saying, we should tell the users to fork the `scalingo-sonarqube` repository, and link the github repo to to their application so that they can trigger a deploy which will update the app. Is that right?

--
Thomas WICKHAM
Business Development Executive @Scalingo
Website: https://scalingo.com
Telephone: +33 6 30 06 23 15
The automagic cloud for high-performing teams - Le cloud automagique pour les équipes performantes

EtienneM:
There is two possibility to update to a more recent version of SonarQube:
1. Wait for the official buildpack to make the new version the new default (https://github.com/Scalingo/sonarqube-buildpack/blob/master/bin/compile#L16) and manually trigger a new deployment.
2. Update the `SONARQUBE_VERSION` environment variable of your application and manually trigger a new deployment.

mackwic:
How do we check if there's update to the `SONARQUBE_VERSION` variable to do ? How often will we update ?

EtienneM:
> How do we check if there's update to the `SONARQUBE_VERSION` variable to do ? 

You need to monitor SonarQube update. I don't know if they have an email notification system or thing like that.

> How often will we update ?

Probably upon user request.

mackwic:
What about using `curl https://binaries.sonarsource.com/Distribution/sonarqube/ | ggrep -Po ' sonarqube-.+zip' | sort | tail -n 1` ? :)

Soulou:
> What about using `curl https://binaries.sonarsource.com/Distribution/sonarqube/ | ggrep -Po ' sonarqube-.+zip' | sort | tail -n 1` ? :)

Non repeatable build, with same env, same code, same buildpack, you've different results → bad practice

mackwic:
What's worse between shipping old software and non-repeatable build ? I'm pretty worried about the updates of this buildpack...

EtienneM:
This is what we do for all our buildpacks. getting rid of repeatable build is really something I wouldn't get rid of. having up-to-date software is the responsibility of the client, not ours. It's like saying: this application has a deprecated library installed, Scalingo should make sure to install a not deprecated one. 

Scalingo just install what you want, the version you want.

mackwic:
Well then, it needs to be written and how to check the build-act installs the last version so that users can effectively takes responsibility.

--
Thomas WICKHAM
Business Development Executive @Scalingo
Website: https://scalingo.com
Telephone: +33 6 30 06 23 15
The automagic cloud for high-performing teams - Le cloud automagique pour les équipes performantes

curzolapierre (approved):
So, LGTM :)

PR #1126, opened 2020-12-08
Update certificate authority url

ksol:


_posts/databases/mongodb/2000-01-01-compass.md
```diff
@@ -62,7 +62,7 @@ Host" form:
 {% include mdl_img.html %}
 
 The "Certificate Authority" is downloadable
-[here](https://db-api.scalingo.com/api/ca_certificate).
+[here](https://db-api.osc-fr1.scalingo.com/api/ca_certificate).
```
EtienneM, line 65: @Soulou don't we want a region-less domain name for such things? This is the exact same file on all regions.

Soulou: Yes and no, in a way that I think it's better that there would be one link per region. because they might got different over time, even if today they are identical.

EtienneM: OK so let's keep it that way for now :)

EtienneM (approved):

PR #1142, opened 2020-12-29
Adding a section about nextjs dynamic port

cyrille-sg:
Adding a section about nextjs dynamic port (reopened on a better page)


EtienneM (commented):
Why don't you add a new page for Next.js? That's what is asked by Luca in https://github.com/Scalingo/one-stop-shop/issues/223

cyrille-sg:
Giving what was asked by Lucas, I was unsure if we will not reorganize a bit the pages. My goal was to add the quickly the note, not thinking about the reorganization / new pages

EtienneM (approved):

PR #1149, opened 2021-01-04
[PHP] Composer default version is 2.x

EtienneM:
Related to Scalingo/php-buildpack#168


EtienneM:
Replaced with https://github.com/Scalingo/documentation/pull/1280

PR #1158, opened 2021-01-06
Deploy a web-less application

### LINKED ISSUE #930: Explain how to run web-less applications
Some customers are interrested in running applications not exposed to the web; while it's possible it's not explained in the doc nor is it easy to guess how (having a dummy `web` container scaled at 0)

EtienneM:
Fix #930


ksol (commented):
Ok with the page, maybe add one or two example use-cases for running webless ?

ksol:
Hm, hosting only workers-like processes on scalingo?

ksol (approved):

PR #1164, opened 2021-01-21
Update slugignore documentation

nclavaud:
Fixing the PHP apps section ([Webpack](https://webpack.js.org/), not [Webpacker](https://github.com/rails/webpacker)), and other minor rewording.


EtienneM (approved):
Thanks a lot! 👍🏾

PR #1204, opened 2021-03-15
GMP is now accessible for PHP 8

Soulou:
Since we've rebuilt all dependencies of PHP, the `gmp` extension now build successfuly for PHP 8

It seems it was a problem of version of the `libgmp` which got updated now.


EtienneM (approved):
Can you also update the status of the related issue: https://github.com/Scalingo/php-buildpack/issues/189

Soulou:
Done

PR #1207, opened 2021-03-18
Update curator version

yalsicor:
There is a dependency error problem with version 5.8.1:

https://github.com/elastic/curator/issues/1560
https://github.com/elastic/curator/issues/1496

This is fixed in the new version 5.8.2


EtienneM (approved):
Thanks :)

PR #1213, opened 2021-03-26
docs(changelog): new API route for one-off container stop

EtienneM:
Do not merge before we deploy the API in prod

Related to [OPSB-255]

[OPSB-255]: https://scalingo.atlassian.net/browse/OPSB-255


Soulou (commented):
I would turn the naming different.

Add an endpoint to stop a container

Asynchronously stops a running container → ONLY one-off can be stopped that way. (It's about the semantic of the endpoint)

Soulou (approved):

PR #1296, opened 2021-05-25
Update composer versions

johnsudaar:


EtienneM (approved):
Thanks, I updated our procedure so that we don't forget to modify this page at the next update

PR #1300, opened 2021-05-27
Fix stack reference in Puppeteer installation docs

yanjost:
Fix heroku-18 stack reference in Puppeteer installation docs


EtienneM (commented):
Are you sure it works? Our base image is currently based on Ubuntu 18.04 like the equivalent `heroku-18` stack. I'm not sure it can work with `heroku-20` in the long term

EtienneM:
> May it be the order of the buildpacks in the .buildpacks file ?

Order of buildpacks in the .buildpacks file is indeed important. That might have been what did the trick in your case :)

PR #1313, opened 2021-06-02
chore: bump ruby version

ksol:
@EtienneM or @Soulou can I leave you checking that the docker-compose is still working good ?


Soulou:
@ksol fix on dockerfile added

ksol:
@Soulou re: bundler, if we can bump to 2.x, I'm all for it. I didn't do it to ease up the bump as I can't debug correctly when running via docker

EtienneM (commented):
When I  execute `dc run --rm web bundle install` I have the error:
```
╰─$ dc run --rm web bundle install
Traceback (most recent call last):
        2: from /usr/local/bin/bundle:23:in `<main>'
        1: from /usr/local/lib/ruby/2.7.0/rubygems.rb:296:in `activate_bin_path'
/usr/local/lib/ruby/2.7.0/rubygems.rb:277:in `find_spec_for_exe': Could not find 'bundler' (1.17.3) required by your /usr/src/app/Gemfile.lock. (Gem::GemNotFoundException)
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:1.17.3`
```

I'm not sure how to solve this.

There are also some paths containing `/usr/src/app/vendor/.bundle/ruby/2.5.0` in the Dockerfile and the `docker-compose.yml` file which should be replaced with `/usr/src/app/vendor/.bundle/ruby/2.7.0`

Soulou:
Please update to 2 then ;-)

ksol:
@EtienneM I think Leo's changes solves your issues. I've bumped to bundler 2 as well, so latest commit _should_ work

EtienneM (approved):

PR #1320, opened 2021-06-18
Add changlog and quick start entry for go 1.16 buildpack

TheoEb:


EtienneM (approved):
You just forgot to update the `modified_at` field, otherwise LGTM

PR #1321, opened 2021-06-25
dashboard changelog entry

ksol:


Soulou (approved):
As @EtienneM is off, I'm doing the review

PR #1335, opened 2021-07-02
docs(elixir): rephrase explanation about community buildpack

EtienneM:
Do you think this wording is better to explain community buildpack?


ksol (approved):

ksol:
I do :)

PR #1359, opened 2021-08-10
changelog entry: links in my.

ksol:


ksol:
I was not sure on the title, feel free to suggest better

Soulou (commented):
Maybe we want to mention `osc-secnum-fr1` also?

ksol:
It's mentionned ! 

> The rules also apply to the `osc-secnum-fr1` dashboard.

PR #1381, opened 2021-09-06
Specific next.js solution

guillemfondin:
Because the solution to my problem was not on the link indicated by the error


guillemfondin:
Ok I see, I posted the same PR twice because I didn't know which page it was the most relevant on :)

PR #1382, opened 2021-09-06
Specific next.js solution

guillemfondin:
Because the solution to my problem was not on the link indicated by the error


EtienneM (approved):
Thanks 👍🏾

PR #1393, opened 2021-09-14
docs(erlang): downgrade some versions

EtienneM:
The OTP we use has been dowgraded in https://github.com/HashNuke/heroku-buildpack-elixir-otp-builds/commit/c9d4faf8a2920fd43111a206b77beb6e29a77552. Let's only use the OTP versions described in https://github.com/HashNuke/heroku-buildpack-elixir-otp-builds/blob/master/otp-versions as per our internal documentation. It seems thos not in the `optp-versions` file have been porrly packaged


curzolapierre (approved):
LGTM

PR #1399, opened 2021-09-16
docs(scala): Specify that Scala should be updated when use a more recent JDK version

brandon-welsch:


_posts/languages/scala/2000-01-01-start.md
```diff
@@ -33,6 +33,15 @@ The buildpack will detect your app as Scala if it has the
 populated `.ivy/cache` into your container. The `.ivy2` directory will be
 cached between builds to allow for faster build times.
 
+## JDK Compatibility
+
+Be careful when you update the Java runtime of your application.
+If the Scala version does not support the version of Java you specified
+the build can most likely fail.
```
EtienneM, line 40: Is the error message in the deployment logs understandable? If it is kind of blurry, can you provide an example of such message here?

brandon-welsch: The message is related to the deployed project. The sample I was able to test, shows an error on a java class not found which is related to the project. I have not put the error here as it is not a generic error.

EtienneM: OK 👍🏾

EtienneM (approved):

PR #1426, opened 2021-10-08
Pngquant got a breaking change

cyrille-sg:
Forcing resolution for now to make it pass https://github.com/imagemin/imagemin-pngquant/issues/46

Forcing resolution will froze the version. As a side effect, any vulnerability fix will not be installed.
Alternative is to add the apt buildpack and install the libpng-dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
     "svgo": "^1.0.4",
     "svgo-loader": "^2.1.0"
   },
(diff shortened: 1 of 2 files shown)


cyrille-sg:
It is not working, as it seems our ubuntu-20 do not onboard the lib png

EtienneM (approved):

PR #1467, opened 2021-11-24
fix(puppeteer): Use new repository for buildpack

mattboll:


EtienneM (commented):
@Soulou do we want to have an official fork of a Puppeteer buildpack? The current buildpack mentioned in the documentation does not exist anymore (https://github.com/levups/puppeteer-scalingo-buildpack).

Soulou:
@EtienneM That's a legit question, we give no real support to our buildpack, it may be interested to mention it.

In our documentation, maybe we want to suggest people to use APT buildpack and give them the list of packages to install in order to run puppeteer?

https://github.com/Scalingo/puppeteer-buildpack/blob/master/bin/compile#L34

Soulou:
@EtienneM or maybe fix the link and keep advertise for https://github.com/yrambler2001/puppeteer-scalingo-buildpack which is maintained it seems

mattboll:
Hi, To help you in the decision, I have to say that I forked the scalingo repo to allow scalingo-20 (and I'm ready to open a PR if you want to keep the repo) I didn't try the repo from yrambler2001 to see if it is up to date and work out of the box

EtienneM:
I think that my favorite solution is:
- Archive the repository https://github.com/Scalingo/puppeteer-buildpack. We don't really want to maintain it, even though the burden seems light.
- Stop mentioning the community buildpack (https://github.com/yrambler2001/puppeteer-scalingo-buildpack) as we have no idea whether it will still be maintained in the future. Just like the buildpack (https://github.com/levups/puppeteer-scalingo-buildpack), it may disappear in the near future.
- Update the documentation to explain how to install Puppeteer with the apt buildpack. The Puppeteer buildpack does not seem to do anything else than installing some apt packages.

Would you agree on this @Soulou?

Soulou:
> 

LGTM, but I think it would be great to have somewhere a list of buildpacks from the community, that people could try without any warranty.

That being said it's probably out of scope of this. So ok with you for now

EtienneM:
Thanks for the suggestion of modifications. A more long-term solution seems to be the one described in https://github.com/Scalingo/documentation/pull/1480

PR #1472, opened 2021-12-01
Outscale Object Storage Unit service is END OF SUPPORT AND MAINTENANCE

vonox7:
See https://docs.outscale.com/en/userguide/About-Object-Storage.html


EtienneM (commented):
Outscale OSU (Object Storage Unit) is end of life, but Outscale OOS (OUTSCALE Object Storage) is still alive :)

PR #1503, opened 2022-01-06
SSH: Switch to ed25519 keys

johnsudaar:
Since DSA and RSA are slowly being rolled out of client implementations


EtienneM (approved):
Did you made sure that this key type is correctly handled in all Scalingo places? If yes, LGTM :)

johnsudaar:
> Did you made sure that this key type is correctly handled in all Scalingo places?

Yes it's the key i'm using for 6 months now.

PR #1532, opened 2022-01-25
docs(postgresql): explain idle queries

### LINKED ISSUE #1198: [PostgreSQL] Explain idle queries
> It's just an open connection from the connection pool

> Usually it means that the client did not close the connection:

```
Each row in pg_stat_activity represents an established connection to the server from a client. "idle" means the client is not currently executing a query nor in a transaction.

If query_start_date is 2 days old, that just means the last query to be executed on that connection was two days ago.

Not really a reason for worry, unless you have so many connections open that it's consuming more RAM than you can afford. It's generally desirable for a connection pool to have a few idle connections so queries don't suffer the latency of establishing a new connection.
```

EtienneM:
https://documentation-service-pr1532.osc-fr1.scalingo.io/databases/postgresql/start#running-queries

I'm not completely convinced by how the quote of the mailing list is displayed. WDYT? Should I paraphrase instead of quoting to improve the display? Or leave it as is?

Fix #1198

### DIFF
--- a/_posts/databases/postgresql/2000-01-01-start.md
+++ b/_posts/databases/postgresql/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: PostgreSQL
 nav: Introduction
-modified_at: 2021-04-27 00:00:00
+modified_at: 2022-01-25 00:00:00
 tags: databases postgresql addon
 index: 1
 ---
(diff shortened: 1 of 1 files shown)


brandon-welsch (approved):
I think the display it's okay. Otherwise what would be the other display?

LGTM

brandon-welsch:
Oh, there's the auto-merge 😄. But i'm fine with that. I just wish I could have seen the other display

EtienneM:
> Otherwise what would be the other display?

I had no idea :D

PR #1557, opened 2022-02-18
fix: Docker file

btrd:


Dockerfile
```diff
@@ -11,10 +11,11 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
 RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
 
 RUN apt-get update
+RUN apt-get install -y python
```
EtienneM, line 14: Out of curiosity, what do we need Python for? 😲

EtienneM (approved):

Dockerfile
btrd: Seems like an NPM package need it, didn't investigate further to be honest

PR #1563, opened 2022-02-28
Update default stack in 2000-01-01-stacks.md

nicolasleger:


nicolasleger:
Also might need to move default disclaimer for https://doc.scalingo.com/platform/internals/stacks/scalingo-18-stack and https://doc.scalingo.com/platform/internals/stacks/scalingo-20-stack

EtienneM (approved):
Thanks

PR #1570, opened 2022-03-09
docs(ffmpeg): install FFmpeg via the APT buildpack

EtienneM:
Related to https://github.com/Scalingo/ffmpeg-buildpack/issues/4


curzolapierre (approved):
Out of curiosity why do you spell `ffprobe` in lowercase, while `FFmpeg` has capital `F`?
From the documentation FFmpeg seems to be the product, ffmpeg and ffprobe are binaries (very nitpicking).

LGTM

EtienneM:
> Out of curiosity why do you spell `ffprobe` in lowercase, while `FFmpeg` has capital `F`?

@curzolapierre This page (https://ffmpeg.org/ffprobe.html) spells with two lower case `f`, that's why I spelled it that way 🤷🏾

curzolapierre:
I agree, but this page https://ffmpeg.org/ffmpeg.html spells ffmpeg with two lower case `f` too :sweat_smile:

PR #1574, opened 2022-03-10
docs(heroku-compatibility): Fix GitHub query, use HTTPS on every links and make the text more airy

brandon-welsch:


EtienneM (approved):
I don't really like creating one paragraph for each sentence. It's seems grammatically odd. I mean, a paragraph can contain multiple sentences. Why would you change that to have more paragraphs, but with a single sentence in each paragraph?

I can validate this PR but I don't really understand the reason of it

brandon-welsch:
I find that in some places in the documentation some things is really stuck together.
It's less readable I think and when there are big chunks it doesn't really make you want to read.

There are places where I agree with having several sentences in a paragraph.
But there are places where it's too much.

brandon-welsch:
I have done some changes. Maybe it's better now

EtienneM (approved):

PR #1576, opened 2022-03-11
docs(mongo-dump-restore): Improve mongo dump and restore page

brandon-welsch:


EtienneM (approved):
I still don't understand all the new lines you added in this file. It won't change anything to the final output

brandon-welsch:
Here I wrap the text, so as not to exceed 80 characters per line. Some lines are much too long.
I know that this does not change the final content.

PR #1591, opened 2022-03-17
oos addon

cyrille-sg:


yohann-bacha:
is this still relevant @cyrille-sg ?

cyrille-sg:
I would say no under the current form, but most of the documentation could be useful to someone having a s3 compatible object storage. So with a bit of refactor, it may be interesting for our users

PR #1595, opened 2022-03-21
Update node default version

btrd:


EtienneM (approved):
Oh nice, I'll update our procedure to mention this

PR #1637, opened 2022-05-10
fix(elascticsearch): need to reach the support to restore a backup

curzolapierre:
We don't propose the restore button anymore, it is needed to reach the support to be able to restore a backup on ElasticSearch.


EtienneM:
> We don't propose the restore button anymore, it is needed to reach the support to be able to restore a backup on ElasticSearch.

Shouldn't we add the restore button again instead of updating the doc? Is there a reason why the button is not displayed anymore?

PR #1667, opened 2022-06-16
fix(scalingo-scheduler): add 15 min timeout information

curzolapierre:


src/_posts/platform/app/task-scheduling/2000-01-01-scalingo-scheduler.md
```diff
@@ -83,6 +83,7 @@ For example if your task run during 5 minutes, you will be billed 5 minutes of a
 ## Long-running Tasks
 
 Scheduled tasks are meant to execute short running tasks.
+There is a timeout set at 15 minutes, after this time the container will be killed.
```
EtienneM, line 86: What do you think about this suggestion? Otherwise LGTM

```suggestion
Tasks are automatically killed after 15 minutes.
```

EtienneM (approved):

PR #1679, opened 2022-07-05
fix: text won't break and enlarges box on mobile screen

Anouchka-M:
[DOC-88]
Allows text to break on small screens.
Long wordings without space as DATABASE_URL=$SCALINGO_MONGO_URL in src/_posts/platform/app were enlarging their own box and text went outside of parents box.

[DOC-88]: https://scalingo.atlassian.net/browse/DOC-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/styles/index.postcss.css
+++ b/src/app/styles/index.postcss.css
@@ -194,6 +194,10 @@ article code.language-plaintext {
   @apply bg-sc-gray-3 text-sc-gray-2 font-mono px-2 py-1 rounded;
 }
 
+body {
+  @apply break-all sm:break-normal;
+}
+
 .highlight .hll { background-color: #060C20; }
 .highlight .c { color: #75715e } /* Comment */
 .highlight .err { color: #960050; background-color: #1e0010 } /* Error */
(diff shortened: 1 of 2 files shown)


src/app/styles/index.postcss.css
```diff
@@ -194,6 +194,10 @@ article code.language-plaintext {
   @apply bg-sc-gray-3 text-sc-gray-2 font-mono px-2 py-1 rounded;
 }
 
+body {
+  @apply break-all sm:break-normal;
```
btrd, line 198: I would only apply this on the content of the page (not on the menu, header, footer, ...).

btrd:
Adding Cyrille has a reviewer, we both work on the documentation together.

src/app/styles/index.postcss.css
cyrille-sg: At first I was not a fan (it is a not specific enough change / it can bleed a lot / we do not have visual tests to ensure no side effect). Like Jean said, I wish to have it set on a more specific target, for the rest I am considering it could the change needed, but I would like to be sure it has been extensively tested, especially by resizing the windows from mobile to 2xl screens

btrd (approved):

src/app/styles/index.postcss.css
Anouchka-M: Moved it to article and corrected the readme with Brandon and Pierre's informations !

PR #1695, opened 2022-07-25
docs(compass): update instructions to match new fields name

EtienneM:


curzolapierre (approved):
LGTM

brandon-welsch:
@EtienneM I don't see why you updated this documentation because the previous fields are correct with the latest version of Compass.
image

PR #1710, opened 2022-08-04
docs(documentation): update Laravel documentation

### LINKED ISSUE #1680: Improve PHP Laravel documentation according to feedbacks
Allright, I took a look at it (great feedback by the way) and here my thoughts regarding the "Laravel integration" part, so far:

the composer.json issue is still relevant, but, IMO, it's more related to the way Composer works than to Scalingo. The documentation could still be a bit more precise/fixed.
I'm not sure I fully understand the issue with the database connection parameters but : according to the Laravel documentation, and starting with Laravel 5.8, the DATABASE_URL env var is fully supported by Laravel, which means:

- there is no need to parse the connection string with Laravel >= 5.8 (it seems like the documentation has been updated in this regard).
- there is no need to map it to the url config option (Laravel will automatically do it)
(I personnaly think that explicit is better than implicit, so I would keep the url mapping in the documentation, but that's arguable).
(also, do we want to re-add documentation for Laravel < 5.8?)
- the Procfile documentation seems OK to me. If this feature is used quite often with Laravel, the suggestion to have an example of a full file could be indeed relevant. Maybe we could add some links to other pages (e.g. the Laravel Queue and Scheduler sections)?
- the documentation now specifies that the container must be scaled to 1 to start the queue. A similar sentence is still missing for the scheduler.

This feedback led me to another thought: when a user creates an app and link it to a public GitHub/GitLab repo, would it be possible to analyze the code and provide the user some links to the relevant pages of the documentation? (e.g. we detect it's a Laravel app --> guide the user to the PHP page, to the Laravel page, to the DB pages and to the logs page, if we detect a Procfile , put a link to the Procfile documentation page, etc.)I haven't checked the others parts of the feedback for now but I will do it ASAP.Please let me know if I'm misunderstanding something or if you are expecting something else :) (edited)

Frzk:
The section of the Laravel documentation dedicated to the Laraval Task Scheduler
is outdated.

This patch fixes this by encouraging users to use the Scalingo
Scheduler. The old way to proceed is still documented for cases where
tasks need to run more frequently than once every 10 minutes.

Fixes #1680

### DIFF
--- a/src/_posts/languages/php/2000-01-01-laravel.md
+++ b/src/_posts/languages/php/2000-01-01-laravel.md
@@ -218,23 +218,46 @@ of your application:
 queues: php artisan queue:work --queue=high,default
 ```
 
-The arguments of the `--queue` option must be customized to suit your needs. After deploying your
-application, you must scale the newly created type of containers `queues` at least to 1.
+The arguments of the `--queue` option must be customized to suit your needs.
+After deploying your application, you must scale the newly created type of
+containers `queues` at least to 1.
 
 ## Laravel Tasks Scheduler
 
-The Laravel PHP framework includes a command scheduler for asynchronous tasks. However, Laravel's
-[documentation](https://laravel.com/docs/8.x/scheduling#introduction) states you need to add a cron
-entry which is not possible on Scalingo.
+The Laravel PHP framework includes a command scheduler for asynchronous tasks.
+However, Laravel's [documentation](https://laravel.com/docs/9.x/scheduling#running-the-scheduler)
+states you need to add a cron entry which is not possible on Scalingo.
 
-{% note %}
-The cron feature is now available upon request. Refer to [this page]({% post_url platform/app/task-scheduling/2000-01-01-scalingo-scheduler %}) for more information.
-{% endnote %}
+From this point, there are two possibilies, depending on the frequency of the
+scheduled tasks.
 
-This documentation page guides you to let your application use the Laravel command scheduler on
-Scalingo. The following command must be added to your app. It calls the Laravel scheduler every
-minutes. It requires you to start [another process]({% post_url platform/app/2000-01-01-procfile %})
-in your application.
+### The Tasks Run on a Frequency Lower Than Once Every 10 Minutes
+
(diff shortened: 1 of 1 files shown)


src/_posts/languages/php/2000-01-01-laravel.md
```diff
@@ -218,23 +218,46 @@ of your application:
...
-This documentation page guides you to let your application use the Laravel command scheduler on
-Scalingo. The following command must be added to your app. It calls the Laravel scheduler every
-minutes. It requires you to start [another process]({% post_url platform/app/2000-01-01-procfile %})
-in your application.
+### The Tasks Run on a Frequency Lower Than Once Every 10 Minutes
+
+For these cases, we highly recommend to use the [Scalingo Scheduler]({% post_url platform/app/task-scheduling/2000-01-01-scalingo-scheduler %}).
+
+All you have to do is create a `cron.json` file at the root of your application
+with a content similar to this:
```
EtienneM, line 239: We usually avoid wording which suggests that something is simple or obvious. That's why we don't use the word "just" or "simply" in the doc. And I have the feeling that the wording "All you have to do" is that kind of construct. But may be just me being picky, as you wish :) 

```suggestion
Create a `cron.json` file at the root of your application with a content similar to this:
```

src/_posts/languages/php/2000-01-01-laravel.md
```diff
@@ -279,73 +302,42 @@ class SchedulerDaemon extends Command
 }
 ```
 
-Don't forget to add the following line to the `Procfile` of your application:
+In the example above, you can see that the defined scheduler runs every minute.
+(If you look closely, you will also see that our scheduler makes use of the one
+provided by Laravel, `schedule:run`.)
+
+You now have to tell Scalingo how to start this new scheduler. To do this,
```
EtienneM, line 309: I have the feeling that "tell" (https://www.wordreference.com/enfr/tell) is not exactly the word to use 🤔 

```suggestion
You now have to instruct Scalingo how to start this new scheduler. To do this,
```

EtienneM (approved):
Just a couple of English question but LGTM

PR #1711, opened 2022-08-04
feat(dependabot): Monthly + btrd assigned to review as owner of the project

EtienneM:
This project has plenty of security issues due to outdated Node.js dependencies. This configuration aims at fixing them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,11 @@ updates:
       interval: "monthly"
     reviewers:
       - "btrd"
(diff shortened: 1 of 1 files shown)


ksol:
We usually don't want to enable dependabot before doing a pass of dependency catch-up. In this case we don't really have tests, and the js ecosystem is at build time more than runtime, so the risks are something we can live with until we have some time to dedicate to this

EtienneM:
OK let's close this then. Thanks for the explanation :)

PR #1755, opened 2022-09-28
specified explicit href

yannski:


yannski:
> Good catch ! I was about to notify you

I surveille ça like the milk on the fire

yannski:
@yanjost can you approve?

yanjost:
If I may, can you change "perimeter" to "scope" ? Perimeter seems to relate only to a topological context

yannski:
> If I may, can you change "perimeter" to "scope" ? Perimeter seems to relate only to a topological context

I use your own words :)

yanjost:
Still not working...

PR #1781, opened 2022-10-21
docs: improve docs on login

lionelB:
For new user, it's not clear that one needs to run `scalingo login` to login 
The actual doc presents `scalingo login` command like it is only for a use with token.


btrd:
Thanks for this change, indeed it's easier for user to simply use `scalingo login`.

btrd (approved):

PR #1804, opened 2022-11-07
Update 2000-01-01-celery-flask.md

huynhdoo:
Update the celery call for the worker since celery version 5


EtienneM (approved):
Thanks for the fix 👍🏾

Issue #1813, opened 2022-11-17
Check work-break css rules to ensure not word cut in the middle


cyrille-sg:
Words can be cut in the middle on mobile, look at css rules.

![Image](https://user-images.githubusercontent.com/62012562/202405705-2c658720-12c5-4250-9a68-eb43b512ee83.png)

https://scalingo.slack.com/archives/C6GE4JP3L/p1668547697196919

Anouchka-M:
@btrd This issue is related to this PR : https://github.com/Scalingo/documentation/pull/1679 and this issue : https://scalingo.atlassian.net/browse/DOC-88

There is a css rule on `article` element which applies `word-break: break-all` on small screens (under 640px).

Problem is that with any other breaking word rule, the content goes outside its box when a word is quite long (and we have some, with env variable by example).

You can see an example with `word-break: break-word`, a rule which theorically should break word when necessary. Unfortunatly it's deprecated and doesn't do the work anymore... https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
![image](https://user-images.githubusercontent.com/94181724/204507633-794cc8f0-b296-434a-bf95-8690c0c6ff50.png)

So we have three options that I can see:
- We don't break words but content goes outside its box
- We keep on breaking words, not that good but it doesn't break the whole page
- We track each "long word" and try to find a solution for it not to break the page (adding a space after an =, using code blocks instead of code markdown formatting... https://www.markdownguide.org/basic-syntax#code)

I'm in favour of the second one, because I guess a majority of our clients doesn't use screens under 640px to read our documentation. This solution also doesn't breaks the whole page (instead of the first one) and won't take us time.

The third solution would be the most "user friendly" but it would require not to forget about it each time we update the documentation.

btrd:
> We track each "long word" and try to find a solution for it not to break the page (adding a space after an =, using code blocks instead of code markdown formatting... https://www.markdownguide.org/basic-syntax#code)

We could apply `word-break: break-all` on the `code` HTML element for small screen, it's use for code block and inline.

btrd:
Do you have examples ? How many letters are needed ? I think there isn't much long work in English.

Anouchka-M:
@btrd We have a problem with tables also, if it can't break in a cell (and the room for a word is quite small).

![image](https://user-images.githubusercontent.com/94181724/204568894-31e878c8-6c10-41c4-801b-1b430c082e2f.png)

Anouchka-M:
[...]

PR #1824, opened 2022-11-30
fix: small screen words breaking, table & code too large

Anouchka-M:
#1813

### DIFF
--- a/src/_posts/platform/internals/stacks/2000-01-01-stacks.md
+++ b/src/_posts/platform/internals/stacks/2000-01-01-stacks.md
@@ -12,12 +12,14 @@ well-known Linux distribution.
 Scalingo currently supports two stacks based on Ubuntu. The corresponding base
 Docker images are open source and can be found on the Docker hub:
 
-{: .table }
+<div class="overflow-horizontal-content" markdown="1">
 | Name | Base Distribution | Supported Until | Docker Hub |
 | ---- | ----------------- | --------------- | ---------- |
 | [scalingo-20]({% post_url platform/internals/stacks/2000-01-01-scalingo-20-stack %}) | Ubuntu 20.04 | April 2025 | [scalingo/scalingo-20](https://hub.docker.com/r/scalingo/scalingo-20) |
 | [scalingo-18]({% post_url platform/internals/stacks/2000-01-01-scalingo-18-stack %}) | Ubuntu 18.04 | April 2023 | [scalingo/scalingo-18](https://hub.docker.com/r/scalingo/scalingo-18) |
 | [scalingo-14]({% post_url platform/internals/stacks/2000-01-01-scalingo-14-stack %}) | Ubuntu 14.04 | December 2019 | [scalingo/scalingo-14](https://hub.docker.com/r/scalingo/scalingo-14) |
+{: .table }
+</div>
 
 All Scalingo buildpacks officially supported by Scalingo are compatible with
 all stacks.
--- a/src/app/styles/index.postcss.css
+++ b/src/app/styles/index.postcss.css
@@ -122,7 +122,7 @@ article table {
   @apply border border-sc-gray-3 table-auto w-full mb-4 text-sc-text-4 text-sc-gray-1;
 }
 
-article .table-wrapper {
+article .overflow-horizontal-content {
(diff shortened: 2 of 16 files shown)


Anouchka-M:
@btrd @guillaume-sc a draft so you may check the suggestion described in the issue comment

src/app/styles/index.postcss.css
```diff
@@ -122,6 +122,10 @@ article table {
   @apply border border-sc-gray-3 table-auto w-full mb-4 text-sc-text-4 text-sc-gray-1;
 }
 
+article .table-wrapper {
```
guillaume-sc, line 125: The `table-wrapper` identifier is too generic IMHO. It doesn't convey the idea its content is allowed to overflow horizontally. If we end up having multiple configurations for wrapping enclosed tables, we'll hardly differentiate them if the first one is simply named "wrapper".

I suggest to name it after the behavior it introduces: `overflow-horizontal-content`, or `overflow-horizontal-content-for-XXX` where XXX relates to `w-[95vw]`.

btrd: What if we apply this to all table inside article ? It simplify things

btrd:
@Anouchka-M Yes I deleted it because the answer was in the issue in your last comment

src/app/styles/index.postcss.css
guillaume-sc: _Disclaimer: I haven't browsed the current source code. Please comment, and disregard my suggestion if it doesn't follow the current code style, or convention._

This is my suggestion, indeed, so it doesn't add one level of indirection for the reader before understanding what it is. The class name `table-wrapper` gives me the place in the HTML code, however at the moment I read it I'm already in the HTML code. It gives me the identifier to look for in the CSS, though this is the indirection I was mentioning.

Anouchka-M: @btrd What do you think about this solution and about the class naming ?

Anouchka-M:
Some `table` elements are on plain html, and others on markdown.
Had to found a solution to add a `div` and apply the class on markdown ones. Found something but I think it's not really "clean" to mix html and md like this.
I also don't understand why we have some `table` on md, and others on plain html.

src/_posts/platform/internals/stacks/2000-01-01-stacks.md
```diff
@@ -12,12 +12,14 @@ well-known Linux distribution.
...
 Docker images are open source and can be found on the Docker hub:
 
-{: .table }
+<div class="overflow-horizontal-content" markdown="1">
 | Name | Base Distribution | Supported Until | Docker Hub |
 | ---- | ----------------- | --------------- | ---------- |
 | [scalingo-20]({% post_url platform/internals/stacks/2000-01-01-scalingo-20-stack %}) | Ubuntu 20.04 | April 2025 | [scalingo/scalingo-20](https://hub.docker.com/r/scalingo/scalingo-20) |
 | [scalingo-18]({% post_url platform/internals/stacks/2000-01-01-scalingo-18-stack %}) | Ubuntu 18.04 | April 2023 | [scalingo/scalingo-18](https://hub.docker.com/r/scalingo/scalingo-18) |
 | [scalingo-14]({% post_url platform/internals/stacks/2000-01-01-scalingo-14-stack %}) | Ubuntu 14.04 | December 2019 | [scalingo/scalingo-14](https://hub.docker.com/r/scalingo/scalingo-14) |
+{: .table }
```
guillaume-sc, line 21: Why does this move _after_ the Markdown table?

guillaume-sc (commented):
I read your concern regarding mixing up Markdown and HTML, however I consider HTML comes to the rescue where Markfown falls short of options.

Can you confirm why there's a `{: .table}` moving after the actual Markdown table? I have no other comment, and I'll approve after you shed light on this point.

src/_posts/platform/internals/stacks/2000-01-01-stacks.md
Anouchka-M: Yes @guillaume-sc , I don't know why but it wasn't working if leaving it up: the "implicit" html under markdown wasn't interpreted so it was just plain text.

guillaume-sc (approved):

PR #1856, opened 2022-12-22
feat: add flags for ssl mongodump and mongorestore

### LINKED ISSUE #1746: [Mongo Dump/Restore] We should add a note to use ssl flags in case of force TLS
When force TLS option is enable we should explain that the dump and restore commands should have ssl parameters:
```
<dump or restore command> --ssl --sslCAFile ca.pem
```

ipfaze:
fix #1746 
Changes can be seen here : 
https://documentation-service-pr1856.osc-fr1.scalingo.io/databases/mongodb/dump-restore#dump-and-restore-with-force-tls


curzolapierre (changes requested):
Please, update `modified_at` field in front matter.
Otherwise LGTM

curzolapierre (approved):

PR #1858, opened 2022-12-26
Add note about Kibana Elasticsearch add-on

aurelien-reeves-scalingo:
The one-click button to deploy Kibana is provisioning a new Elasticsearch add-on. This may be confusing while the documentation indicates how to use the one that has been provisioned just before that with Logstash.


EtienneM (approved):
nice catch 👍🏾

PR #1887, opened 2023-01-10
Add required config for Vite

cyrille-sg:
Add required config for Vite

### DIFF
--- a/src/_posts/languages/nodejs/2000-01-01-start.md
+++ b/src/_posts/languages/nodejs/2000-01-01-start.md
@@ -345,6 +345,6 @@ To fix this issue, modify the `start` script by adding `-p $PORT`
 In the `vite.config` (or `vite.config.ts`) file you need to add the following
 ```
   server: {
-    host: true,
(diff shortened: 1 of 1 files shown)


src/_posts/languages/nodejs/2000-01-01-start.md
```diff
@@ -339,3 +339,12 @@ To fix this issue, modify the `start` script by adding `-p $PORT`
   }
 }
 ```
+
+## Vite
+
+In the `vite.config` (or `vite.config.ts`) file you need to add the following
+```
+  server: {
+    host: true,
```
EtienneM, line 348: From Vite documentation (https://vitejs.dev/config/server-options.html#server-host), I feel like it would be more understandable to use this value rather than `true`:

```suggestion
    host: "0.0.0.0",
```

Moreover, won't we also have an issue with the port (https://vitejs.dev/config/server-options.html#server-port)?

EtienneM (approved):

PR #1904, opened 2023-01-23
  Fix deploy on scalingo

btrd:
Fix deploy on scalingo


btrd:
Replace by https://github.com/Scalingo/documentation/pull/1905

PR #1912, opened 2023-01-31
Adding warning about review apps

cyrille-sg:


guillaume-sc (changes requested):
Please update some of the wording, as suggested.

guillaume-sc (approved):

PR #1961, opened 2023-02-14
docs(PostgreSQL): warn that DB is down during PiTR restore

### LINKED ISSUE #1959: [PostgreSQL] Point-in-time recovery needs the database to be down
https://doc.scalingo.com/databases/postgresql/start#point-in-time-recovery

It is not mentioned that the use of this feature requires the database to be stopped.

EtienneM:
Fix #1959


yohann-bacha (approved):
LGTM

PR #2005, opened 2023-03-01
db-tunnel command needs an ssh setup

SCedricThomas:


btrd:
Ping @SCedricThomas, should this be close or merge?

SCedricThomas:
I hesitate to add a note in several places, but it should not be a problem if we just mention it once

btrd (approved):

PR #2017, opened 2023-03-10
docs(CD): dpl is considered deprecated

### LINKED ISSUE #1987: [GitLab CI with DPL] Job marked as OK instead of failure
I followed the documentation to integrate Scalingo within Gitlab CI (using dpl).

The main problem of this tool seems to be that when an error is triggered during the build (ex: buildpack job failed) the job is considered as OK, which is ... dangerous.

When browsing dpl repository, it is not clear if the project is actively maintained (everything is alpha supported, since more than 6 months, low activity on repo).

A a Scalingo customer, I would see no objection of removing dpl if considered as a deprecated way of deploying. But ideally, scalingo would provide a clear and simple way to use scalingo CLI in gitlab runner environment...
Or Scalingo could provide support for dpl to ensure everything is working fine.

Thanks in advance !

Example of log :
```shell
!     An error occurred during buildpack compilation
 !   Error deploying the application
!   → Invalid return code from buildpack
To ssh.osc-fr1.scalingo.com:vera-production.git
   21620e6..b8e69ea  HEAD -> master
$ timeout 60 ./scalingo keys-remove dpl_tmp_key
Key 'dpl_tmp_key' has been deleted.
Cleaning up project directory and file based variables
00:00
Job succeeded
```

EtienneM:
We tried opening PR to update the Scalingo dpl provider with no news for months. Let's consider dpl as deprecated. We should not mention deprecated / barely working solution to deploy on Scalingo. It brings support and is confusing.

Fix #1987


brandon-welsch (approved):
LGTM

PR #2019, opened 2023-03-10
changelog(deployment): Increase default image size to 1.5GiB

SCKevinO:


src/_posts/platform/deployment/2000-01-01-limits.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Deployment Limits
 nav: Limits
-modified_at: 2018-08-10 00:00:00
+modified_at: 2023-03-10 14:35:00
```
brandon-welsch, line 4: FYI The time is not required to be changed, it's useless. But the date is required 😉

brandon-welsch: Could you also now update the time before merging?

src/changelog/deployment/_posts/2023-03-10-default-image-size-1-5gb.markdown
```diff
@@ -0,0 +1,8 @@
+---
+modified_at: 2023-03-10 14:35:00
+title: 'New default image size limit: 1.5GiB'
```
brandon-welsch, line 3: Same here, please update 🙏🏻

brandon-welsch (approved):
LGTM

src/changelog/deployment/_posts/2023-03-10-default-image-size-1-5gb.markdown
```diff
@@ -0,0 +1,8 @@
+---
```
brandon-welsch, line 1: @SCKevinO Can you also change the filename which depend on the timestamp too?

brandon-welsch (changes requested):

src/changelog/deployment/_posts/2023-03-10-default-image-size-1-5gb.markdown
brandon-welsch: And i'm not sure the date currently match when we sent in production

brandon-welsch (approved):
LGTM

PR #2021, opened 2023-03-13
Added GIT_FETCH_EXTRA_FLAGS variable

sauerben:
Still head the Shallow Error with the `GIT_DEPTH` variable in our case. Found this issue and it worked with `GIT_FETCH_EXTRA_FLAGS: --unshallow` https://gitlab.com/gitlab-org/gitlab/-/issues/292470

Maybe consider adding it.


sauerben:
After some deployments, this is causing some other issues. So please ignore the PR.

PR #2023, opened 2023-03-14
Note on runtime and secret file

cyrille-sg:
We often say that one-off are a copies of what is deployed, but there are some caveats.

### DIFF
--- a/src/_posts/platform/app/2000-01-01-secret-file-in-app.md
+++ b/src/_posts/platform/app/2000-01-01-secret-file-in-app.md
@@ -45,3 +45,5 @@ $HOME/bin/run
 
 Commit both files and deploy your application. Your application has now access
 to your secret file in `/app/private.pem`!
+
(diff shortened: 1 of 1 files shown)


src/_posts/platform/app/2000-01-01-secret-file-in-app.md
```diff
@@ -45,3 +45,5 @@ $HOME/bin/run
 
 Commit both files and deploy your application. Your application has now access
 to your secret file in `/app/private.pem`!
+
+As a note, the Procfile being executed at runtime, in this case the `web` runtime, the `/app/private.pem` file would not be visible in a one-off. As a one-off being a copy of the deployable environnement, but without what a web runtime execution would add.
```
EtienneM, line 49: suggestion: I understand what you mean but I don't find the paragraph easily readable. 
```suggestion
Note that the file `/app/private.pem` is not available in a one-off container. This solution makes it available when booting the `web` container. Hence the file is only available in `web` containers and not one-off containers.
```

EtienneM (approved):
I'm proposing another wording. Feel free to use it or not. I'm already validating the PR.

btrd:
Ping @cyrille-sg

PR #2045, opened 2023-03-28
docs(wkhtmltopdf): add warning to mention deprecation

### LINKED ISSUE #1550: Wkhtmltopdf is not maintain anymore
https://doc.scalingo.com/platform/deployment/buildpacks/wkhtmltopdf

We should add a warning on this page and also propose an alternative.

EtienneM:
Fix #1550


btrd:
Is it our buildpack that is deprecated or wkhtmltopdf? Because from what I'm seeing it's neither 🤔

PR #2074, opened 2023-04-07
Update Rails redis cache timeout configuration examples

Intrepidd:
Rails sets a default `connect_timeout` of 20 : https://github.com/rails/rails/blob/89fc4bf9196f861372a8cb0daae59f415789403f/activesupport/lib/active_support/cache/redis_cache_store.rb#L55-L60

Setting `timeout` when the `connect_timeout` is set has no effect on it, easily testable by running this : 

```
Redis.new(connect_timeout: 20, timeout: 1, url: "redis://10.0.0.0:9876").get("foo")
```

So I'm proposing to update the config example so it doesn't suffer this issue, is more explicit and showcases all options

Thanks !


EtienneM:
@btrd I'm assigning the review to you as you are more a Rails expert than I am

btrd:
@Intrepidd Thank you very much for your review, it completely make sense!

btrd (approved):

PR #2111, opened 2023-05-23
docs(mongodb): wording + grammatical issues

### LINKED ISSUE #954: [MongoDB] Plan upgrade operation should include more info about downtime
A few things I would like to change in this section (https://doc.scalingo.com/databases/mongodb/start#database-upgrade): 
- "This operation happens instantly thanks to Docker containers": remove mention of Docker containers. It's confusing and useless
- More informatino about downtime during upgrade. For instance, an upgrade from starter to business (without changing the amount of RAM) induces no downtime (we just add new instances). Maybe explain more the process, like Léo did on Slack:

> All long would take the whole operation?

It depends of the size of the database: after each instance of the cluster is restarted, we're waiting it gets back to a healthy state (SECONDARY), and not a transient state, the duration of this changes according the size of the DB. idea of the evolution: 1GB -> seconds, 10GB -> 10s seconds -> 100GB -> minutes

> Is there any data transfers when I scale vertically ?

At some point the current servers are not able to cope with the load of the databases present on it. But the migrations are not done during the plan change, as the plan change usaully doesn't trigger a change of usage of the host usage. But later when we get alerts from our monitoring related to the overloading of a node, we might migrate data from one node to another. This is done manually today, and can be done almost transparently. Almost because sometimes we need to trigger a failover of the primary if we need to migrate it.

> What’s the expected impact for the app in term of latency peak or throughput ?

Any operation requiring a failover leads to a latency peak during the operation, a failover in Mongo means the following: current node announces it's not primary anymore, and triggers an election it won't answer to let another node becoming primary, during this time, no write/read operation is allowed, connections are supended. It lasts between 3secs and 10 secs usually. The logics about reconnect and retry is 100% in the clients. (MongoDB has developed "clever" clients for all main technologies)so tl;dr; a change of plan triggers a latency of 3s <-> 10s, normally no request should fail but it depends of the client.

EtienneM:
Actually stalled regarding the situation with Mongo Inc.

Fix #954

### DIFF
--- a/src/_posts/databases/mongodb/2000-01-01-start.md
+++ b/src/_posts/databases/mongodb/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: MongoDB
 nav: Introduction
-modified_at: 2022-04-28 00:00:00
+modified_at: 2023-05-23 00:00:00
 tags: databases mongodb addon
 index: 1
 ---
@@ -135,7 +135,9 @@ mongo --ssl --sslCAFile=/path/to/ca.pem "<connection string>"
 
 ## Changing Plans
 
-You can upgrade or downgrade your database plan whenever you need it. This operation happens instantly thanks to Docker containers and no manual input is required. When you change the plan, your database will be stopped then restarted on a new host with new parameters of the chosen plan. During the operation the connection is dropped between your app and the database. Finally, after the operation is successful, the related app will be restarted.
+You can upgrade or downgrade your database plan whenever you need it. This operation happens with no manual input. The impact of changing the plan depends on what plans we are talking about:
+* When you change the amount of RAM available to your database, your database will be stopped then restarted on a new host with new parameters of the chosen plan. During the operation the connection is dropped between your app and the database. Finally, after the operation is successful, the related app will be restarted.
+* When you change from starter to business (keeping the amount of RAM), there is no downtime as only new nodes are added.
 
 ### From the Dashboard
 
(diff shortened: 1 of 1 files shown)


yohann-bacha:
As the issue is closed, but this PR is fixing grammatical issues, should we merge it @EtienneM ?

src/_posts/databases/mongodb/2000-01-01-start.md
```diff
@@ -135,7 +135,9 @@ mongo --ssl --sslCAFile=/path/to/ca.pem "<connection string>"
 
 ## Changing Plans
 
-You can upgrade or downgrade your database plan whenever you need it. This operation happens instantly thanks to Docker containers and no manual input is required. When you change the plan, your database will be stopped then restarted on a new host with new parameters of the chosen plan. During the operation the connection is dropped between your app and the database. Finally, after the operation is successful, the related app will be restarted.
+You can upgrade or downgrade your database plan whenever you need it. This operation happens with no manual input. The impact of changing the plan depends on what plans we are talking about:
```
yohann-bacha, line 138: **nitpick:** "we" is a bit unformal, and "what" is not well suited here

```suggestion
You can upgrade or downgrade your database plan whenever you need it. This operation happens with no manual input. The impact of changing the plan depends on which plans are involved:
```

src/_posts/databases/mongodb/2000-01-01-start.md
```diff
@@ -135,7 +135,9 @@ mongo --ssl --sslCAFile=/path/to/ca.pem "<connection string>"
...
+* When you change the amount of RAM available to your database, your database will be stopped then restarted on a new host with new parameters of the chosen plan. During the operation the connection is dropped between your app and the database. Finally, after the operation is successful, the related app will be restarted.
```
yohann-bacha, line 139: **nitpick:** Let's breathe a bit, and I think that we must not use the future?

```suggestion
* When you change the amount of RAM available to your database, your database will be stopped then restarted on a new host with new parameters of the chosen plan. During the operation, the connection is dropped between your app and the database. Finally, after the operation is successful, the related app is restarted.
```

yohann-bacha (approved):
i just left some nitpicks, LGTM

PR #2165, opened 2023-06-20
Fix documentation about router log

btrd:


EtienneM (commented):
I don't understand this PR. Yes router logs are separated from application logs. They have different retention policy. For legal purpose we need to keep the router logs for more times than the application logs.

btrd:
We may keep the router logs, but they're not accessible to customers if the `Router logs setting isn't enabled.

Reading this paragraph, the customer was thinking he could access the router log elsewhere, but that's not the case. The router log setting is not a setting "to mix router logs into your application logs", it's a setting to make router logs accessible to customers or not.

See level 3 tickets https://www.notion.so/scalingooriginal/Access-router-log-when-not-enable-047eafefbdf9441187fbcb31267cc101?pvs=4

And Intercom conversation https://app.intercom.com/a/inbox/w4oogu7s/inbox/shared/all/conversation/12375700176295?view=List

EtienneM (approved):
Thanks fr the explanation :)

Issue #980, opened 2020-06-09
Databases: explain the users/databases/rights


ksol:
It's not always clear what rights have the databases users, which db/schema they can read, and why they can't do some operations (read some schemas, add more databases, etc).
 
imo, there should be a page for each supported DB technology that first explains at a high level the concepts, then gives the technical-oriented details

EtienneM:
The PostgreSQL details have been added in #537

EtienneM:
oh wow, thanks Ksol 🤩 👋

PR #2168, opened 2023-06-22
docs: documents users rights for MySQL, Redis and Elasticsearch

EtienneM:
Related to #980


src/_posts/databases/redis/2000-01-01-start.md
```diff
@@ -270,6 +270,15 @@ Beware that no downgrade is possible once your database has been upgraded.
   </table>
 </div>
 
+### Database Users
```
curzolapierre, line 273: Nitpick: I wonder if it wouldn't more relevant to place this section under the `Redis Configuration` one.
Because I think it is more related to the configuration of the database instead of the dashboard since it is not possible to create a new user from it.

curzolapierre (commented):
Just a comment, otherwise LGTM

src/_posts/databases/redis/2000-01-01-start.md
EtienneM: I wanted this section to be at the exact same place on all database types doc pages. That's why I wrote it here. But if you think it should be moved I can. What do you prefer? :)

PR #2172, opened 2023-06-28
Replicate from source to target

jeremybdk:
According to the documentation here:
https://developer.redis.com/riot/#_replicate

The first URL is the source, the second the target.


EtienneM (approved):
Thanks for the fix :+1:

PR #2208, opened 2023-07-26
fix(databases): specify to setup ssh for db-tunnel

### LINKED ISSUE #1984: Add a mention to the SSH key setup guide in the tunneling example
We should mention that the SSH key setup is mandatory to create an db tunnel.
We should mention this [URL](https://doc.scalingo.com/platform/getting-started/first-steps#ssh-key-setup) for each tunnel setup example.

curzolapierre:
Fixes #1984


EtienneM (approved):
Just a non-blocking suggestion, but already approving the PR

PR #2211, opened 2023-07-27
Docs/#2210/db migrations postdeploy

### LINKED ISSUE #2210: [Ruby] Explain the best practices for DB migrations
The explanation of how to handle database migration after a deployment is lacking in our documentation.

We frequently have this question on support.
It can be a sensitive subject, because a bad DB migration can be destructive.

We should explain why we don't automatically execute DB migration as other PaaS do, cf: https://app.intercom.com/a/inbox/w4oogu7s/inbox/shared/all/conversation/12375700102761
A nice explanation has already been given, for example: https://app.intercom.com/a/inbox/w4oogu7s/inbox/conversation/12375700059658

Anouchka-M:
Fixes  #2210

### DIFF
--- a/src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
+++ b/src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
@@ -161,16 +161,13 @@ To run a migration, you need to use a [post-deployment hook]({% post_url platfor
 
 The Ruby application starts up before the schema has changed, but does not receive a query.
 
-As a result, if operations are started before the migration has been executed, the application may load the schema `pre-migration`, and so a restart may be necessary after the migration, however short it may be.
+As a result, if operations are started before the migration has been executed, the application may load the schema `pre-migration`. A restart could be necessary after the migration.
 
 The best practice is to perform non-destructive migrations, where a column rename is performed in several steps, rather than an `ALTER TABLE` renaming a column.
 
 A good example in `strong_migrations` gem [documentation](https://github.com/ankane/strong_migrations#renaming-a-column).
 
-{% note %}
-  Running a migration on an application that then crashes on startup can be risky, so the application must start without the database migration.
-  As we make sure the application starts before running the hook, if it fails, `postdeploy` won't be run and database migration won't be executed either.
-{% endnote %}
+{% include app_start_without_migration.md %}
 
 ## Where to go next
 
--- a/src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
+++ b/src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
@@ -1,6 +1,6 @@
 ---
 title: Get Started with Ruby on Scalingo
-modified_at: 2018-02-23 00:00:00
+modified_at: 2023-07-27 00:00:00
 tags: ruby sinatra tutorial getting-started-tutorial
 index: 3
 ---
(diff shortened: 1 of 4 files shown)


src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
```diff
@@ -155,6 +155,23 @@ $ git push scalingo master
  <-- https://my-ruby-app.osc-fr1.scalingo.io -->
 ```
 
+## Database migrations
+
+To run a migration, you need to use a [post-deployment hook]({% post_url platform/app/2000-01-01-postdeploy-hook %}).
+
+The Ruby application starts up before the schema has changed, but does not receive a query.
+
+As a result, if operations are started before the migration has been executed, the application may load the schema `pre-migration`, and so a restart may be necessary after the migration, however short it may be.
```
EtienneM, line 164: I find this sentence too long, hence hard to understand. Can you slice it in smaller pieces?

src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
```diff
@@ -155,6 +155,23 @@ $ git push scalingo master
...
+
+The best practice is to perform non-destructive migrations, where a column rename is performed in several steps, rather than an `ALTER TABLE` renaming a column.
+
+A good example in `strong_migrations` gem [documentation](https://github.com/ankane/strong_migrations#renaming-a-column).
+
+{% note %}
+  Running a migration on an application that then crashes on startup can be risky, so the application must start without the database migration.
+  As we make sure the application starts before running the hook, if it fails, `postdeploy` won't be run and database migration won't be executed either.
```
EtienneM, line 172: suggestion: it's unclear to me what does "we" refers to
suggestion: shorter == easier to understand sentence

```suggestion
  Scalingo makes sure the application starts before running the hook. If the application does not start, the post-deployment hook won't run. Consequently the database migration won't be executed either.
```

src/_posts/languages/ruby/rails/2000-01-01-start.md
```diff
@@ -120,6 +120,23 @@ $ git commit -m "Configure application for Scalingo"
 $ git push scalingo master
 ```
 
+### Database Migration
+
+To run a migration, you need to use a [post-deployment hook]({% post_url platform/app/2000-01-01-postdeploy-hook %}).
+
+The Ruby on Rails application starts up before the schema has changed, but does not receive a query. The way ActiveRecord works on Rails is that the schema is loaded into memory on the first query on a given table.
+
+As a result, if initializers, for example, or an asynchronous worker start performing operations before the migration has been executed, the application may load the schema `pre-migration`, and so a restart may be necessary after the migration, however short it may be.
```
EtienneM, line 129: The part "if initializers, for example, or an asynchronous worker" is too long and makes the sentence complicated to read.

This is a really long sentence that I'm not sure to understand. I don't understand the "however short it may be" sentence.

EtienneM (changes requested):
Thanks, it was missing for a long time from our documentation! Some small comments but overall LGTM :)

You copy-pasted the same content in two different places. In such cas, you can use the `include` feature of Jekyll. You can refer to the Pierre's PR #2208 to see how it works. If you don't understand I can help you do it

src/_posts/languages/ruby/2000-01-01-getting-started-with-ruby.md
Anouchka-M: @EtienneM Thanks to you about the review !

The content is not exactly the same, as there are specific mentions to Rails who aren't present in the Ruby page, but the note part is. I'll apply that to the note.

src/_posts/languages/ruby/rails/2000-01-01-start.md
Anouchka-M: Ok, trying to make it clearer !

curzolapierre (commented):
I wonder if we really want to details in the getting started how to execute migration.
I think we don't want to go into too much detail in `getting started`.
At least, a mention to the rails guide (containing more details) should be enough, WDYT?

Anouchka-M:
@curzolapierre The problem I see is user who don't use Rails but are on Ruby and maybe other frameworks, such as this user : https://app.intercom.com/a/inbox/w4oogu7s/inbox/admin/5676128/conversation/12375700182825?view=List

I think users on Rails will check directly Rails documentation, and users on other Ruby frameworks will check Ruby documentation. It could be extracted in a separated file under Ruby folder maybe ?

EtienneM (approved):

PR #2249, opened 2023-08-28
docs: VPN addons are only for runtime

### LINKED ISSUE #1662: [OpenVPN] Mention that the VPN connection is done during runtime and not build time
https://doc.scalingo.com/addons/scalingo-openvpn/start

EtienneM:
Here is how the block looks:

Fix #1662


ipfaze (commented):
Would it be more visible inside a `warn` block?
WDYT?

ipfaze:
Yep, you're right a warn block is a bit too strong.
A note block seems fair :slightly_smiling_face:

ipfaze (approved):
Perfect :slightly_smiling_face:

PR #2255, opened 2023-08-30
docs(datadog): give an example of .buildpacks content

EtienneM:
A customer got confused and didn't noticed it's needed to also add the language buildpack. I think the page was slightly unclear about that.


ipfaze (approved):
LGTM :+1:

PR #2277, opened 2023-09-11
fix(documentation): improve Metabase documentation

### LINKED ISSUE #2246: Metabase documentation should mention database replication
We should mention that we do not recommend connecting Metabase to the production database.
Consequently, we should also mention how to duplicate the prod database to another one to be used by Metabase.

Linked to #2245

Frzk:
Reworded some paragraphs and added some insights about Metabase deployment.
Fixes #2246

### DIFF
--- a/src/_posts/platform/getting-started/2000-01-01-getting-started-with-metabase.md
+++ b/src/_posts/platform/getting-started/2000-01-01-getting-started-with-metabase.md
@@ -21,9 +21,11 @@ under 5 minutes.
   for this purpose.
 
 - Depending on several factors such as the amount of data stored in your
-  production database, its load and the complexity of the Metabase queries you
+  production database, its load, and the complexity of the Metabase queries you
   want to run, you may consider duplicating your production data to an
-  additional database, dedicated for Metabase use. [Our documentation]({% post_url platform/databases/2000-01-01-duplicate %})
+  additional database dedicated for Metabase use. Doing so would prevent
+  Metabase to have a negative impact on your application's performances.
+  [Our documentation]({% post_url platform/databases/2000-01-01-duplicate %})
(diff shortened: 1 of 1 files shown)


src/_posts/platform/getting-started/2000-01-01-getting-started-with-metabase.md
```diff
@@ -1,53 +1,89 @@
...
-## Metabase Deployment
+## Deploying Metabase
+
+### Planning your Deployment
+
+- Metabase requires its own database to store its configuration and some metadata.
+  We usually advise to use a [PostgreSQL Starter/Business 512 addon](https://scalingo.com/databases/postgresql)
+  for this purpose.
+
+- We also strongly advise against directly plugging Metabase to your production
```
Soulou, line 23: I'm not really agreeing on this.

* If database is small (< few GB), a read-only user is really good enough (performance and security wise)
* If database is larger, then a copy is required to prevent impacting the production database performance since BI queries may be costful.

Soulou (changes requested):

src/_posts/platform/getting-started/2000-01-01-getting-started-with-metabase.md
Frzk: Ok :+1: 
Would it be possible to have a better estimation for "< few GB"? :smiley:

Soulou: Like hmmm ~5GB

Frzk: I finally chose another, more generic, wording.
Would that suit?

Soulou (approved):

PR #2323, opened 2023-10-23
Update S3 instructions for WP

SwiTool:
- Resource can be an array, avoiding duplicate rules.
- Added instructions on bucket policy, as by default, every object is not public, even is the bucket is.


SwiTool:
`s3:DeleteObject` permission was added because Medias can be deleted by an admin in wordpress. It also deletes the file in the bucket.

EtienneM (approved):
Thanks a lot :+1:

PR #2326, opened 2023-10-24
Update 2000-01-01-setup-ssh-macos.md

smohadjer:
The instructions for setting up ssh on mac missed this section!


EtienneM (approved):
Thanks! I also added the section "Check everything is working".

PR #2352, opened 2023-11-03
Update CircleCI default config

vala:
The existing CircleCI config seems outdated and didn't work without the proposed changes for me with CircleCI config 2.1.

Hope this can help others to integrate Scalingo with their CI.


EtienneM (approved):
Great, thanks a lot! :clap:

PR #2374, opened 2023-11-29
feat(changelog): release of NodeJS 20.10.0 and 21.2.0

Frzk:
Fix https://github.com/Scalingo/nodejs-buildpack/issues/96

### DIFF
--- a/src/_posts/languages/ruby/rails/2000-01-01-start.md
+++ b/src/_posts/languages/ruby/rails/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: Get Started with Ruby on Rails on Scalingo
 nav: Get Started
-modified_at: 2023-11-29 12:00:00
+modified_at: 2023-07-27 00:00:00
 tags: ruby rails tutorial getting-started-tutorial
 index: 1
 ---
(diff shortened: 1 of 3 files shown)


src/_posts/languages/ruby/rails/2000-01-01-start.md
```diff
@@ -149,7 +149,7 @@ but it's expected. There is nothing in the project, it's time to build your prod
 
 Latest version of Rails application also use Node.js in case ExecJS or
 Webpacker is detected. The Node.js version built inside the Ruby buildpack is
-frozen. It is the latest version of the LTS 16.x at the time of writing.
+frozen. It is the latest version of the LTS 20.x at the time of writing.
```
EtienneM, line 152: issue: this is not correct. The Ruby buildpack embeds a simple support of Node.js and in this context v16.x is installed by default:

https://github.com/Scalingo/ruby-buildpack/blob/2aba1378b09270d817189dd1368fbd66344db36f/lib/language_pack/helpers/nodebin.rb#L5

EtienneM (changes requested):

src/_posts/languages/ruby/rails/2000-01-01-start.md
Frzk: Right, I was anticipating the `ruby-buildpack` update :sweat_smile:
Should be reverted now :)

Issue #1777, opened 2022-10-13
External links with anchor get google-analytics-id added to anchor


vonox7:
E.g. https://doc.scalingo.com/platform/app/child-apps has a link at the bottom of the page ([here](https://developers.scalingo.com/deployments.html#trigger-manually-a-deployment-from-an-archive&_ga=2.12345678.1234567654.1234567-12345678.132435467)).

But the link contains the `&_ga=xxx` value at the end, which breaks the anchor-link:

Original anchor:
`#trigger-manually-a-deployment-from-an-archive`
Real anchor:
`#trigger-manually-a-deployment-from-an-archive&_ga=2.12345678.1234567654.1234567-12345678.132435467`

cyrille-sg:
Thanks for the issue, we can reproduce it, we're gonna work on a fix

aurelien-reeves-scalingo:
I still reproduce the issue using Chromium , or Firefox with ad-blocker and advanced ad-tracker protection disabled
The link seems good at first, but the google tag id is added once the link has been clicked

aurelien-reeves-scalingo:
Nice catch!

Could you ask in #ask-tech-team who is responsible for the configuration of GA, then ask to take a look at that configuration?

leo-scalingo:
Google Analytics has been disabled in the documentation since the version installed does not work anymore, I'll check with Jennifer how to improve things, closing for now, the bug is not present anymore

PR #2445, opened 2024-01-24
Fix url for metabase deploy link

clemlatz:
Current link to deploy metabase does not work:

Capture d’écran 2024-01-24 à 11 46 42

Using link from https://github.com/Scalingo/metabase-scalingo/blob/master/README.md fixes this:

Capture d’écran 2024-01-24 à 11 47 51


EtienneM:
This is actually due to #1777, not because of the URL.

PR #2480, opened 2024-02-12
fix(erlang): remove Erlang from documentation

Frzk:
Fix https://github.com/Scalingo/project-items/issues/559

I'm not sure if we want to make something about Elixir?


ipfaze (approved):
LGTM :+1:

PR #2481, opened 2024-02-13
feat(changelog): release of CLI 1.30.1

curzolapierre:
Related release: https://github.com/Scalingo/cli/releases/tag/1.30.1


EtienneM (commented):
I would use the content of the changelog file that seems more useful to our customers (https://github.com/Scalingo/cli/blob/master/CHANGELOG.md#1301). What do you think?

And can you ping people on the issue https://github.com/Scalingo/cli/issues/1029 to let them know that release 1.30.1 fixed the issue?

curzolapierre:
I followed what we did before, but I agree with you.

Yep it was planned :+1:

EtienneM (approved):

PR #2487, opened 2024-02-27
Update 2000-01-01-restoring.md

SebouChu:
Use correct environment variable to restore backup (based on previous comand).


Frzk (approved):
Indeed, thanks a lot!

SebouChu:
Thanks for the speed ;)

PR #2528, opened 2024-03-29
Update 2000-01-01-setup-ssh-macos.md

alanzirek:
Proposing an addition to the documentation to explain configuring ~/.ssh/config for users with multiple SSH keys


EtienneM (approved):
Thanks a lot for your contribution. I took the liberty to slightly update your PR but it's really useful. Thanks.

PR #2529, opened 2024-03-29
Update url to get an API token from Datadog

clemlatz:
The url to get an API token from Datadog is invalid


leo-scalingo (approved):

leo-scalingo:
Thanks for the contribution

PR #2561, opened 2024-04-26
[STORY-443] Update documentation to mention releases and how to create domain without Let's Encrypt

leo-scalingo:


curzolapierre (approved):
LGTM

PR #2605, opened 2024-05-03
fix: Fix database duplicate example script

nclavaud:
Hi there,

I have been using the example script on this page to copy my database (thanks for providing the script!):
https://doc.scalingo.com/platform/databases/duplicate

Overall the script is fine, but I ran into a small issue: when retrieving the backup filename from the TAR archive, we need the filename part *after* the leading slash (second field) and not the one before (first field). Hence we should use `cut -f 2` instead of `cut -f 1`.

```sh
$ tar --list --file=backup.tar.gz  | tail -n 1 | cut -d "/" -f 2
tar: Removing leading `/' from member names
20240503001956_*****************.pgsql
```


Frzk (approved):
You're indeed right, thanks a lot for the fix <3

Issue #2621, opened 2024-05-22
H2 and H3 titles are difficult to distinguish


baptmaury:
### [A Review App is available on the PR to test the difference with the prod]

At the moment, the `h2` (`24px`) and the `h3` (`22px`) are difficult to distinguish.

![Capture d’écran 2024-05-22 à 17 31 28](https://github.com/Scalingo/documentation/assets/148763329/d2d6a450-7870-40d6-a7c1-86917cc2fef0)

**I propose to switch `h2` to `26px`.**

PR #2622, opened 2024-05-22
feat(css): Update h2 font-size to make documentation more readable

baptmaury:
Set h2 to 26px

ticket: https://github.com/Scalingo/documentation/issues/2621

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -96,14 +96,14 @@ module.exports = {
         },
       ],
       "sc-title-2": [
-        "24px",
+        "26px",
         {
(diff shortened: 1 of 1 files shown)


EtienneM (approved):
I always felt the same as you, difference between h2 and h3 were too close. Thanks for this change, I really appreciate it :+1:

baptmaury:
> I always felt the same as you, difference between h2 and h3 were too close. Thanks for this change, I really appreciate it 👍

I don't think it's the best, we can make the title even more distinctive, but I don't know if we want to spend that time

PR #2663, opened 2024-06-18
feat: changelogs for nodejs-buildpack and python-buildpack

Frzk:
Fix https://github.com/Scalingo/nodejs-buildpack/issues/124
Fix https://github.com/Scalingo/python-buildpack/issues/101


EtienneM (approved):
:tada:

PR #2665, opened 2024-06-19
docs(mysql|postgresql): add command to connect when in one-off

### LINKED ISSUE #2562: MYSQL Command
https://doc.scalingo.com/databases/mysql/accessing#using-a-one-off

I think it would be interesting to go all the way here and put in the MYSQL command (mysql -u USERNAME --host HOST --port PORT --ssl-mode REQUIRED -p) to be able to connect to the database.

I've just had a person on support  this kind of problem.

EtienneM:
image

Fix #2562

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: '2'
 services:
   # Replicates the prod setup, in which puma serves the file built by jekyll + the assets
   web:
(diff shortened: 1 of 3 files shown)


Frzk (commented):
If we want to keep things a bit homogenized, I suggest we also add a paragraph in the PostgreSQL documentation. The "issue" with this is that PG doesn't come with a single binary so it's "hard" to advise something meaningful :-/

Otherwise, LGTM, thanks :heart_hands:

Frzk (approved):
praise: nice and simple

PR #2670, opened 2024-06-20
docs(account): explain password update with GitHub login

### LINKED ISSUE #1976: Password with GitHub login
Related to this conversation : 
https://scalingo.slack.com/archives/C03T384J3/p1676988428291169

It would be nice to add information about how and why to add a password with GitHub login.
In some cases, a client using his GitHub login needs to add a password (login with the CLI, activating Two-factor authentication and maybe other cases) but the "normal" behaviour of the platform in this case isn't explained.

Is the field asking for the previous password present or not ? 
If yes, should the user just click in it to activate it, or ignore it ?

It happened that a client doesn't remember if he added a password in the past, and without knowledge of the "normal" behaviour, he didn't know if he needed to set or reset his password.

EtienneM:
Fix #1976

### DIFF
--- a/src/_posts/platform/account/2000-01-01-manage.md
+++ b/src/_posts/platform/account/2000-01-01-manage.md
@@ -1,6 +1,6 @@
 ---
 title: Manage your Account
-modified_at: 2023-12-22 00:00:00
+modified_at: 2024-06-20 00:00:00
 tags: account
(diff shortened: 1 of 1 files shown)


Anouchka-M (approved):
Great, thanks!

PR #2698, opened 2024-07-10
Update 2000-01-01-getting-started-with-elk.md

sc-david-voisin:
Fixing what I believe to be some typos


EtienneM (approved):
praise: thanks :tada:

PR #2704, opened 2024-07-18
 Adding information on the differences between the Sandbox, Starter, Business

benjaminach:


aurelien-reeves-scalingo (approved):
Please avoid requesting reviews from more than one person unless special condition. It may be disturbing the folks you are requesting the reviews, so we try to limit the disturbance

benjaminach:
OK, I didn't know.
Thanks for the review.

PR #2709, opened 2024-07-24
docs(environment): hostname composed of app name and container type

revolunet:
Hello

Suggestion for the docs : looks like `HOST` is not exactly an alias for `APP` but the full qualified domain name.

### DIFF
--- a/src/_posts/platform/app/2000-01-01-environment.md
+++ b/src/_posts/platform/app/2000-01-01-environment.md
@@ -118,7 +118,7 @@ additional variable `$PORT` is defined.
 * `CONTAINER_SIZE`: Name of the size of the container `M`, `L`, `XL` etc.
 * `CONTAINER_MEMORY`: Available RAM memory of the container (in bytes)
 * `APP`: Name of the application deployed
-* `HOSTNAME`: Alias for `APP`
+* `HOSTNAME`: The full application hostname (ex: `[APP].[region].scalingo.io`)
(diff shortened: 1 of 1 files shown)


EtienneM (approved):
Thanks a lot for raising the fact that the `HOSTNAME` environment variable is not an alias of the application name.

I took the liberty to edit your modification. After double checking on a running application, it seems the `HOSTNAME` is defined with the app name and the container type name.

revolunet:
Thanks

revolunet:
Hi @EtienneM, is there any way to get the deployed full domain name ?

i tried to add `FQDN=$HOSTNAME.$REGION_NAME.scalingo.io` in my environment variables (from the scalingo UI) but its not interpolated in the container. any idea ?

EtienneM:
> Hi @EtienneM, is there any way to get the deployed full domain name ?
> 
> i tried to add `FQDN=$HOSTNAME.$REGION_NAME.scalingo.io` in my environment variables (from the scalingo UI) but its not interpolated in the container. any idea ?

You won't be able to add a new environment variable that is a composition of multiple environment variables. You need to define this variable in your application code. And you should replace `HOSTNAME` with `APP`.

revolunet:
You're right. I think it could be useful tho to have that FQDN in the variables given. Doing it at app level adds some scalingo coupling

EtienneM:
Yes definitely, I'm forwarding this request internally :)

PR #2742, opened 2024-08-06
Update 2000-01-01-compliance.md

SC-Jerome-S:
correcting the URL (ISO 27001 certificate)


EtienneM (approved):
praise: nice catch, thanks :+1:

PR #2748, opened 2024-08-14
feat: Add a section about Daphne in the Django getting started

### LINKED ISSUE #2474: When using the Daphne CLI, the `--proxy-headers` flag should be provided
It is described here: https://github.com/django/daphne/issues/193
It could be a good idea to add some documentation about daphne as it is quite popular lately

SCedricThomas:
fix #2474

### DIFF
--- a/src/_posts/languages/python/django/2000-01-01-start.md
+++ b/src/_posts/languages/python/django/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: Get Started with Django on Scalingo
 nav: Get Started
-modified_at: 2024-02-01 15:00:00
+modified_at: 2024-08-22 00:00:00
 tags: python django tutorial getting-started-tutorial
 index: 1
 ---
--- a/src/_posts/languages/python/django/2000-01-01-start.md
+++ b/src/_posts/languages/python/django/2000-01-01-start.md
@@ -100,6 +100,18 @@ Then update `Procfile`:
 web: bash bin/start.sh
 ```
 
+### Using Daphne
+
+It is also possible to use Daphne as an ASGI server.
+
+Here an example start.sh script to use Daphne:
+
+```sh
+#!/bin/bash
+
+daphne -b 0.0.0.0 -p $PORT myapp.asgi:application --proxy-headers 
+```
+
 ## Application Configuration
 
 The configuration of the application has to be done through the environment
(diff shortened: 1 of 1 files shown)


src/_posts/languages/python/django/2000-01-01-start.md
```diff
@@ -100,6 +100,18 @@ Then update `Procfile`:
 web: bash bin/start.sh
 ```
 
+### Using Daphne
+
+It is also possible to use Daphne as an ASGI server.
```
EtienneM, line 105: suggestion: maybe add a link to the Daphne homepage?

src/_posts/languages/python/django/2000-01-01-start.md
```diff
@@ -100,6 +100,18 @@ Then update `Procfile`:
...
+
+Here an example start.sh script to use Daphne:
```
EtienneM, line 107: issue: 
- the sentence is not really well written. It's really hard to understand. I don't think it's a very good English.
- it's kind of hard to understand you refer to the `start.sh` script from the previous section
- I am not sure it should be a different paragraph than the previous one.
question: How is it possible to install the `daphne` CLI?

```suggestion
It is also possible to use Daphne as an ASGI server. You can update the `bin/start.sh` script described in the previous section to use Daphne to start the application:
```

src/_posts/languages/python/django/2000-01-01-start.md
```diff
@@ -100,6 +100,18 @@ Then update `Procfile`:
...
+
+```sh
+#!/bin/bash
+
+daphne -b 0.0.0.0 -p $PORT myapp.asgi:application --proxy-headers
```
EtienneM, line 112: question: are there long options to replace `-b` and `-p`?

EtienneM (changes requested):
Did you tested on the Django sample that it works?

SCedricThomas:
Here is an example of the django sample using Daphne as AGSI server: https://sample.osc-fr1.scalingo.io/

EtienneM (commented):
Sorry, just one thing I forgot about in the previous review. Can you update the front matter to update the  last modification date of the page? Then it LGTM

SCedricThomas:
Oh right :face_holding_back_tears:

EtienneM (approved):

PR #2755, opened 2024-08-22
docs(nodejs): fix `devDependencies` installation explanation

### LINKED ISSUE #2645: [Javascript] devDependencies documentation is incorrect, they are now installed by default
Our nodejs-buildpack and our NodeJS documentation https://doc.scalingo.com/languages/nodejs/start#install-devdependencies

It seems our documentation does not match reality on the handling of devDependencies.
The buildpack behaviour has changed, devDependencies are always installed, then they are pruned at the end of the build (so only production dependencies are left during the runtime)

Our documentation states the opposite and should be updated.

EtienneM:
Fix #2645

### DIFF
--- a/src/_posts/languages/nodejs/2000-01-01-start.md
+++ b/src/_posts/languages/nodejs/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: Node.js
 nav: Introduction
-modified_at: 2023-11-29 12:00:00
+modified_at: 2024-08-22 00:00:00
 tags: nodejs
 index: 1
 ---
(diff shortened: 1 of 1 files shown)


src/_posts/languages/nodejs/2000-01-01-start.md
```diff
@@ -92,19 +92,11 @@ Buildpack]({% post_url platform/deployment/buildpacks/2000-01-01-ssh-key %}),
...
-variable to false via our web interface or by using our CLI:
+By default, dependencies present in the `devDependencies` field are installed. At the end of the deployment, `devDependencies` are pruned. Hence only production dependencies are left in the image used for the runtime.
 
-```bash
-$ scalingo --app my-app env-set NPM_CONFIG_PRODUCTION=false
-```
-
-After setting the NPM_CONFIG_PRODUCTION environment variable, you have to
-redeploy your code.
+The pruning step can be skipped by setting the environment variable `YARN_PRODUCTION` to any value.
```
mcimbaro, line 99: Sure? Setting `YARN_PRODUCTION` to `true` shouldn't skip

EtienneM: I'm pretty sure, yes:

https://github.com/Scalingo/nodejs-buildpack/blob/5169a57250df80a532f1aa9c4e19b8bb18ed3b46/lib/dependencies.sh#L176-L179

mcimbaro: Seems a bit counter-intuitive, but you're right

mcimbaro (approved):

src/_posts/languages/nodejs/2000-01-01-start.md
EtienneM: Thanks for paying attention to this detail :)

PR #2756, opened 2024-08-22
[STORY-850] Add postgis_sfcgal to allowed extensions list

leo-scalingo:


EtienneM (approved):

EtienneM:
Maybe also add a changelog entry?

Soulou:
@EtienneM you're right

EtienneM (approved):
Nice, thanks :)

PR #2788, opened 2024-09-05
Fix typo in next standalone guide

totakoko:


EtienneM (approved):
Oh nice catch, thanks for this :+1:

PR #2832, opened 2024-10-08
docs(internals): add SSH public keys fingerprints

### LINKED ISSUE #1897: Add SSH key fingerprint somewhere in the documentation
In order to allow users to properly verify SSH info they are about to put in their `known_hosts` file when interacting with scalingo servers (like when pushing some code in order to deploy an app), it could be nice to have those info somewhere in our documentation.

EtienneM:
Fix #1897

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ object(s)`. Then, insert it with:
 To install dependencies locally:
 
 ```sh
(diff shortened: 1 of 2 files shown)


mcimbaro (approved):
Looks good to me!

Issue #2839, opened 2024-10-10
[Node.js] Confusion about how to prune `devDependencies`


EtienneM:
We have confusing information about how to prune the `devDependencies`. 

On the Node.js > Introduction page, we have a section entitled "devDependencies Installation" (https://doc.scalingo.com/languages/nodejs/start#devdependencies-installation):

> The pruning step can be skipped by setting the environment variable `YARN_PRODUCTION` to any value.

On the page "Yarn 2 and Yarn 3 Compatibility", we have a section entitled "Skip Pruning" (https://doc.scalingo.com/languages/nodejs/yarn-2#skip-pruning). But this section also contains instruction for npm. I find it odd. 

Maybe we should move the content of the section "Skip pruning" in the Node.js > Introduction page?

EtienneM:
Fixed by #2841

PR #2841, opened 2024-10-11
[Node.js] Confusion about how to prune devDependencies

TheoPouletSouchard:
See https://github.com/Scalingo/documentation/issues/2839

### DIFF
--- a/src/_posts/languages/nodejs/2000-01-01-start.md
+++ b/src/_posts/languages/nodejs/2000-01-01-start.md
@@ -104,7 +104,7 @@ If you need access to packages declared under devDependencies at runtime or in a
 
 | Package manager | Version | Environment variable    | Value              |
 |-----------------|---------|-------------------------|--------------------|
-| npm             | Any     | `NPM_CONFIG_PRODUCTION` | `true`  or `false` |
+| npm             | Any     | `NPM_CONFIG_PRODUCTION` | `true`  |
 | yarn            | v1      | `YARN_PRODUCTION`       | `true`             |
 | yarn            | v2+     | `YARN2_SKIP_PRUNING`    | `true`             |
 
--- a/src/_posts/languages/nodejs/2000-01-01-start.md
+++ b/src/_posts/languages/nodejs/2000-01-01-start.md
@@ -102,11 +102,16 @@ The pruning step can be skipped by setting the environment variable `YARN_PRODUC
 
 If you need access to packages declared under devDependencies at runtime or in a different buildpack then, depending on the package manager used, you can set one of the following environment variables to skip the pruning step:
 
-| Package manager | Version | Environment variable    | CLI                                                         |
-|-----------------|---------|-------------------------|-------------------------------------------------------------|
-| npm             | Any     | `NPM_CONFIG_PRODUCTION` | `scalingo --app my-app env-set NPM_CONFIG_PRODUCTION=false` |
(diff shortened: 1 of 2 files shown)


TheoPouletSouchard:
@EtienneM I tried something with a table, it might be easier to read, WDYT ? (i can rollback to old way if you think it was better before)

EtienneM (changes requested):
I like the idea of a table, but the CLI column is too much. It does not render well:

image

Maybe remove this column, add a column "Value" and add a line after the table like:

```
scalingo --app my-app env-set variable=value
```
or something like that?

TheoPouletSouchard:
> I like the idea of a table, but the CLI column is too much. It does not render well:
> 
> image
> 
> Maybe remove this column, add a column "Value" and add a line after the table like:
> 
> ```
> scalingo --app my-app env-set variable=value
> ```
> 
> or something like that?

Uhm i see what you mean. I'm wondering: as the values just need to be set and can be any string to work, the "value" column is useless, isn't it ? 

It might be better to:
- Remove the CLI column-
- Add the CLI line under the table (as you suggested) 
- Add a small text to say "the values can be anything as long as they are set to be valid" (written in a better way)

WDYT?

Frzk:
Hello @TheoPouletSouchard

Thanks a lot for this PR <3

> Uhm i see what you mean. I'm wondering: as the values just need to be set and can be any string to work, the "value" column is useless, isn't it ?

- For `NPM_CONFIG_PRODUCTION`, it seems the value MUST be either `true` or `false` (See https://github.com/Scalingo/nodejs-buildpack/blob/a2a95190bcc7482f61d3dc7b4cc853d25785a117/lib/environment.sh#L61)
- For `YARN_PRODUCTION`, it seems it MUST be set to `true` to be effective. (see https://github.com/Scalingo/nodejs-buildpack/blob/a2a95190bcc7482f61d3dc7b4cc853d25785a117/lib/failure.sh#L371)
- For `YARN2_SKIP_PRUNING`, it seems it MUST be set to `true` to be effective. (see https://github.com/Scalingo/nodejs-buildpack/blob/a2a95190bcc7482f61d3dc7b4cc853d25785a117/lib/dependencies.sh#L181)

While I strongly agree that the presence/absence of the environment variable should be enough, it seems it's not the case.

Consequently, I suggest to replace the CLI column of the table with a "Value" column and put an example below.

src/_posts/languages/nodejs/2000-01-01-start.md
```diff
@@ -98,6 +98,21 @@ By default, dependencies present in the `devDependencies` field are installed. A
 
 The pruning step can be skipped by setting the environment variable `YARN_PRODUCTION` to any value.
 
+#### Skip Pruning
+
+If you need access to packages declared under devDependencies at runtime or in a different buildpack then, depending on the package manager used, you can set one of the following environment variables to skip the pruning step:
+
+| Package manager | Version | Environment variable    | Value              |
+|-----------------|---------|-------------------------|--------------------|
+| npm             | Any     | `NPM_CONFIG_PRODUCTION` | `true`  or `false` |
```
EtienneM, line 107: issue: well, from the code sent by François, the value to skip the pruning should be `true`, shouldn't it?

```suggestion
| npm             | Any     | `NPM_CONFIG_PRODUCTION` | `true`  |
```

EtienneM (changes requested):

EtienneM (approved):
Thanks a lot!! :rocket:

PR #2845, opened 2024-10-21
fix(python): update start page

Frzk:
Fix https://github.com/Scalingo/python-buildpack/issues/119

### DIFF
--- a/src/_posts/languages/python/2000-01-01-start.md
+++ b/src/_posts/languages/python/2000-01-01-start.md
@@ -47,8 +47,9 @@ order of precedence):
 
 - The `runtime.txt` file (deprecated)
 - The `.python-version` file (recommended)
-- The `python_full_version` and `python_version` fields in the `Pipfile.lock`
-  file (Pipenv users only)
+- The `python_full_version` field in the `Pipfile.lock` file (Pipenv users
+  only)
+- The `python_version` field in the `Pipfile.lock` file (Pipenv users only)
 
 {% warning %}
 The `runtime.txt` file is deprecated. If your app uses it, we recommend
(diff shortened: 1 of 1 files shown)


src/_posts/languages/python/2000-01-01-start.md
```diff
@@ -22,17 +22,18 @@ The following versions of Python are available:
...
 | `3.9`          | up to `3.9.20`  | up to `3.9.20`  |
 | `3.8`          | up to `3.8.20`  | unsupported     |
 
 {% note %}
 Even though we still support them, we strongly advise against using deprecated
-versions of Python, such as Python `3.7.x`. If you are still using one of these
-deprecated versions, please consider migrating to a more recent version.
+versions of Python, such as Python `3.9.x` and below. If you are still using
+one of these deprecated versions, please consider migrating to a more recent
+version.
```
EtienneM, line 36: suggestion: I understand why the note here, but with the current information in the doc I don't understand how a customer would know what is a "deprecated" version. What do you think about removing the deprecated Python versions from the table above?

EtienneM (approved):
I disabled auto merge. I added a suggestion.

Already approving the PR. Feel free to merge as is if you want

src/_posts/languages/python/2000-01-01-start.md
Frzk: I hesitated ^^
I've removed them, and I've also removed the warning in https://github.com/Scalingo/documentation/pull/2845/commits/eef5bba7df4bb36779e490cf402e0e509d3e3d84

EtienneM: Nice, I like when we are on the same page :D

EtienneM (approved):

PR #2869, opened 2024-11-07
fix(nginx-buildpack): clean old versions

### LINKED ISSUE #2865: [nginx] list of versions is too big
image

_from https://doc.scalingo.com/platform/deployment/buildpacks/nginx#specify-nginx-version_

IMO we should only highlight the versions that are still officially supported.

In the case of Nginx, I think it should only be the LTS, hence it is only 1.26.

What do you think @Frzk ?

Frzk:
Fix #2865


EtienneM (commented):
Awesome, thanks :clap:

EtienneM (approved):

Issue #2843, opened 2024-10-11
Laravel: document the requirement for the `php artisan optimize` command before starting the app


aurelien-reeves-scalingo:
It may not be mandatory, but `php artisan optimize` seems to be pretty important for Laravel apps. We may want to document how to execute that command before starting the app in the dedicated Laravel documentation

for example via a start.sh script, or right in the Procfile: `php artisan optimize && bin/run`

PR #2876, opened 2024-11-14
fix: How to run `php artisan optimize` using postdeploy hook

sc-zenokerr:
Add instructions to laravel page

Fixes issue #2843

### DIFF
--- a/src/_posts/languages/php/2000-01-01-laravel.md
+++ b/src/_posts/languages/php/2000-01-01-laravel.md
@@ -210,29 +210,18 @@ be automatically run.
 
 ## Optimize: php artisan optimize
 
-It is common to call `php artisan optimize` after a deployment to optimize the
-application. 
-
-You can do this by calling the command in the `postdeploy` process type in the
-Procfile:
-
-```yaml
-postdeploy: php artisan optimize
-```
-
-Or, when you have more complex commands to run you can write a short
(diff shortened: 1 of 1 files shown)


aurelien-reeves-scalingo (changes requested):
The `optimize` command requires to be run before app start-up

It won't have any effect on the app start-up when ran in a postdeploy hook

sc-zenokerr:
I think it will have an effect, because as far as I understand it, php reads the files every request. It will read the cache if it is there.
What do you think?

aurelien-reeves-scalingo:
> I think it will have an effect, because as far as I understand it, php reads the files every request. It will read the cache if it is there. What do you think?

this is not how postdeploy hook are working
it is not the right place to run `php artisan optimize`

`php artisan optimize` seems to generate some cache on the filesystem. postdeploy hook has no impact on the app container filesystem.

the command has to be run at build time, or at runtime before we actually start the app server. There is no place that I know to run it at build time, so it has to be run at runtime

postdeploy is neither build time or app runtime. It is a hook that is run after the build, alongside the runtime of the app. It is run as part of a one-off container after it has been built.

sc-zenokerr:
Of course, it won't modify the filesystem of the running app.
The latest push takes into account your comments.

aurelien-reeves-scalingo (approved):

PR #2882, opened 2024-11-27
feat(postgresql): add Upgrading page for PostgreSQL

Frzk:
STORY-1091
This PR creates a new page in Database/PostgreSQL describing the different upgrades and information about them. This should ease the communication with customers when we deprecate a PostgreSQL version and "force" them to upgrade.
I hope this will also be helpful for support.

PR #2885, opened 2024-11-29
fix(postgresql): apply fixes from review

Frzk:
Apply fixes from #2882 review, since it was merged too quickly :disappointed:


Frzk:
Thanks to YOU!

PR #2797, opened 2024-09-20
Add postgresql query progress api

GradedJestRisk:
## Problem
Some features on PG progress API are not supported on Scalingo, but are not documented.
Check [this blog post](https://blog.octo.com/usual-suspects).

## Solution
Add resources using official docs.

PR #2913, opened 2024-12-13
feat(postgresql): update monitoring instructions

Frzk:
Add information about viewing and monitoring queries and processes from the CLI.

Related to #2797


EtienneM (approved):
Neat :rocket:

PR #2931, opened 2024-12-27
Add ruby 3.3.6 in ruby doc

baptmaury:
Add support of ruby `3.3.6`


Frzk:
Fixed in https://github.com/Scalingo/documentation/pull/2932

PR #2952, opened 2025-01-07
fix(cli): add missing `--app` flag for backups commands

Frzk:
Fix missing `--app my-app` flag with the `backups` command.


EtienneM (approved):
Thanks :+1:

PR #2961, opened 2025-01-14
fix(php-buildpack): revamp extensions page

Frzk:
- List native, PECL and proprietary extensions available.
- For PECL and proprietary ones, also list versions.
- Add instructions to enable and add extensions.


src/_posts/languages/php/2000-01-01-extensions.md
```diff
@@ -14,43 +14,143 @@ Some of them are pre-installed and shipped by default with the deployed version
...
+To add a PHP built-in extension that is not available, please get in touch
+with our support team.
+
 
 ## Working With PECL Extensions
 
-All [PECL extensions](https://pecl.php.net/) are available. If the extension
-you need is not in the above list of pre-installed extensions, it will be
-compiled during the *build* phase of your deployment.
+### Available PECL Extensions
```
EtienneM, line 105: I think that _all_ PECL extensions are available, aren't they? This is my understanding of the code here:

https://github.com/Scalingo/php-buildpack/blob/04665141895505dd51aadf83beba9c2f1c49c9f2/lib/pecl

leo-scalingo (approved):
So great, thanks @Frzk

src/_posts/languages/php/2000-01-01-extensions.md
leo-scalingo: Yes right, it's either we have them pre-built or or we try to compile them on the fly

Frzk: @EtienneM: What @leo-scalingo wrote :)

The ones I listed are the one we pre-compile (maybe it's worth mentioning this for a better understanding). For all others, we try to compile them on the fly, which works in most cases.

BUT sometimes it requires more work (because dependencies, etc.).

And sometimes it doesn't work.
- For example I tried to install the `datadog_trace` PECL extension yesterday and couldn't manage to do it :(
- Another example is the `scoutapm` PECL extension, which is in fact a Zend extension and requires to be enabled with specific instructions (we handle it for the users)

Consequently, writing that "_we support all PECL extensions_" is a bit too much IMHO. That's why I decided to remove this sentence. I'll add precisions though, you're right.

I'll also try to steal a bit of @leo-scalingo's time someday to talk about all this (related to https://github.com/Scalingo/php-buildpack/issues/363)

leo-scalingo (approved):
I really like the improvement

src/_posts/languages/php/2000-01-01-extensions.md
```diff
@@ -14,57 +14,195 @@ Some of them are pre-installed and shipped by default with the deployed version
...
+{
+  "require": {
+    "ext-pdo_pgsql": "*",
+    "ext-gettext": "*",
+    ...
+  }
+}
+```
+
+### Adding a PHP Built-in Extension
```
EtienneM, line 103: suggestion(non-blocking):
```suggestion
### Add Support for a New PHP Built-in Extension
```

src/_posts/languages/php/2000-01-01-extensions.md
```diff
@@ -14,57 +14,195 @@ Some of them are pre-installed and shipped by default with the deployed version
...
+| `ext-memcached` | Optional | Optional | Optional | Optional |
+| `ext-mongodb`   | Enabled  | Enabled  | Enabled  | Enabled  |
+| `ext-oci8`      |          | Optional | Optional | Optional |
+| `ext-redis`     | Enabled  | Enabled  | Enabled  | Enabled  |
 
-To enable a native PHP extension, add it in the `require` block of your
-`composer.json`, like so (notice the `ext-` prefix):
+#### Available Versions
+
+The following table lists the supported versions of available PECL extensions:
```
EtienneM, line 149: question: I don't know if the table is the best here. It looks like all extensions have the same version, whatever the PHP version. Or maybe a table that only lists the versions? But adding a column for each PHP version does not seem really useful, what do you think?

EtienneM (approved):
A couple of suggestions but already LGTM. I completely forgot about this one :sweat_smile:

src/_posts/languages/php/2000-01-01-extensions.md
Frzk: I was expecting this question :smiley:

MHO about this:
- Indeed, **currently** all versions are the same. But sometimes it isn't the case, and I guess it won't always be the case. This is a bet on future ^^
- Specifying everything for each supported PHP version allows to remove any doubt (for the record, I've been having doubts while building the tables: "do we _really_ have the same version for PHP 8.1 and 8.4? Or did we miss an update?")
- For extension `oci8` (for example), it felt a bit weird to have a version while it's not available at all for PHP 8.1.
- I agree that it creates big tables with redundant information, and that it will be boring to maintain. But I also think it makes it clearer and more homogeneous, which gets my priority.

WDYT?

EtienneM (approved):

PR #3041, opened 2025-02-21
[STORY-1326] fix(css): fix table view on mobile

baptmaury:
Fix table view on mobile without using 
`<div class="overflow-horizontal-content" markdown="1">`


Frzk (approved):
question: are the `!important` mandatory?

PR #3096, opened 2025-03-31
Update 2025-03-11-SPM-2025-001.md

luis-scalingo:
Change title level for Changelog section.


EtienneM:
@luis-scalingo is it expected that there is no reviewer on this pull request?

luis-scalingo:
I think so, I didn't ask for one, as I passed through the "public" page and was expecting that public pull requests would be handled by some team (UFS?) in some cycle.
Also, it is a 1 character PR 😄 
Could you tell me if I should ask for a request?

PR #3154, opened 2025-05-22
[STORY-1996] Tech - Add codeowners

annabelle-scalingo:
CODEOWNERS file aims to add automatically reviewers for all PR linked to this repo

we need to add a custom config to select only the files concerned by the automation for reviewers assignment

there are discussions about this: 
https://github.com/orgs/community/discussions/159006
https://stackoverflow.com/questions/75531641/exclude-codeowners-from-dependabot-pr

if we target files updated by dependabot : package-lock.json, gemfile.lock, etc. we shouldn't miss any PR but we may be assigned by default to unrelated PR. 
if it's a "manual" PR, the dev can unassign reviewers if necessary
I don't think this can be a problem, and we can follow this 
WDYT?


aurelien-reeves-scalingo (approved):

aurelien-reeves-scalingo:
That's lame :(

Thanks for having digging this

Could you forward the info to #team-tech belongs in your thread (forwarding to the channel too) related to the CODEOWNERS file?

PR #3241, opened 2025-07-08
feat(note blocks): better rendering of Markdown in note blocks

Frzk:
We have a Liquid tag for notes (`{% note %}`), which, unfortunately, doesn't support some Markdown. This is especially relevant for links.

This new version:
- Puts the icon in CSS, where it belongs
- Allows the following syntax:
  ```markdown
  {% note %}
  [link text][link ref]
  {% endnote %}

  ...

  [link ref]: https://example.com
  ```

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,6 +11,7 @@ excerpt_separator: <!--more-->
 
 # Build settings
 markdown: kramdown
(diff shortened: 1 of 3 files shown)


SC-Samir:
Do you have a picture before and after?

Issue #3276, opened 2025-08-06
Document WEB_CONCURRENCY for Python


sc-david-voisin:
As we have per Ruby and PHP, can we have a documentation about `WEB_CONCURRENCY` for Python.

https://doc.scalingo.com/languages/ruby/start#web_concurrency
https://doc.scalingo.com/languages/php/start#php-fpm-concurrency

PR #3294, opened 2025-08-18
Add web_concurrency for Python

CalianasJ:
feature #3276


src/_posts/languages/python/2000-01-01-start.md
```diff
@@ -119,3 +119,18 @@ dependencies. They must be located in the `bin` root folder of your application
 
 More information at
 [https://github.com/Scalingo/python-buildpack](https://github.com/Scalingo/python-buildpack).
+
+## WEB_CONCURRENCY
+
+The level of concurrency configured is defined automatically according to the memory of the containers of your application. If you want to override this value, yuo can define the environment variable: `WEB_CONCURRENCY`.
```
sc-david-voisin, line 125:
```suggestion
The level of concurrency configured is defined automatically according to the memory of the containers of your application. If you want to override this value, you can define the environment variable: `WEB_CONCURRENCY`.
```

sc-david-voisin (commented):
From https://github.com/Scalingo/python-buildpack/blob/master/vendor/WEB_CONCURRENCY.sh#L9 the number of CPUs affects the final default concurrency value.
What about to add a remark on this point ? 
Or even better, what about provide the default concurrency by container size as done for [php-fpm concurrency](https://doc.scalingo.com/languages/php/start#php-fpm-concurrency) ?

PR #3343, opened 2025-09-19
fix: update WEB_CONCURRENCY table

Frzk:
Supersedes #3325

Thanks @goulvench

### DIFF
--- a/src/_posts/languages/ruby/2000-01-01-start.md
+++ b/src/_posts/languages/ruby/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: Ruby
 nav: Introduction
-modified_at: 2025-09-17 00:00:00
+modified_at: 2025-09-19 00:00:00
 tags: ruby
 index: 1
 ---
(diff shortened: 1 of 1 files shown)


src/_posts/languages/ruby/2000-01-01-start.md
```diff
@@ -155,18 +155,25 @@ end
...
+The default level of concurrency is automatically defined, depending on the
+amount of memory available in the container running the app:
+
+| Container Size | Memory (MB) | Default Concurrency |
+| -------------: | ----------: |-------------------: |
+| S              | 256         | 1                   |
+| M              | 512         | 1                   |
+| L              | 1024        | 2                   |
+| XL             | 2048        | 2                   |
+| 2XL            | 4096        | 4                   |
```
EtienneM, line 170: The previous table contained the `8192` memory. Shouldn't we keep it?

src/_posts/languages/ruby/2000-01-01-start.md
```diff
@@ -155,18 +155,25 @@ end
...
+
+For further details about this calculation, please see the
+[`lib/language_pack/ruby.rb`][web_concurrency_file] file of the Ruby buildpack.
```
EtienneM, line 173: question: is there a reason not to include the link directly here? When reading the markdown file I find it clearer to directly see the actual link

```suggestion
[`lib/language_pack/ruby.rb`](https://github.com/Scalingo/ruby-buildpack/blob/master/lib/language_pack/ruby.rb) file of the Ruby buildpack.
```

EtienneM (changes requested):

src/_posts/languages/ruby/2000-01-01-start.md
Frzk: Good question, our official public catalog includes our containers up to 4096 (https://scalingo.com/pricing). That's why I chose to remove the 3XL line from the table.
Let me know if you'd rather want to include it :)

PR #3375, opened 2025-10-03
Update 2000-01-01-start.md

sc-david-voisin:
Propose the PHP_VERSION usage to set PHP version for classic apps.

### DIFF
--- a/src/_posts/languages/php/2000-01-01-start.md
+++ b/src/_posts/languages/php/2000-01-01-start.md
@@ -69,6 +69,9 @@ For example, to install the latest PHP version of the `8.2` branch:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Frzk (approved):
nitpick: I'd probably put the sentence after the note, so that everything related to Composer is in one block.

EtienneM (changes requested):
Can you update the title to match our convention? It must follow the karma convention (https://karma-runner.github.io/6.4/dev/git-commit-msg.html).

EtienneM (approved):
I think the best title would be your commit message:

> docs(php): select version using `PHP_VERSION`

I updated the title accordingly

PR #3474, opened 2025-12-29
docs(redis): update plan change process downtime information

### LINKED ISSUE #3283: Scalingo for Caching > Changing Plans is wrong
The following paragraph is outdated:
```
You can upgrade or downgrade your database plan whenever you need it. This operation happens instantly thanks to Docker containers and no manual input is required. When you change the plan, your database will be stopped then restarted on a new host with new parameters of the chosen plan. During the operation the connection is dropped between your app and the database. Finally, after the operation is successful, the related app will be restarted.
```

> When you change the plan, your database will be stopped
When the plan change from business to business, only a single node is updated at a time, ensuring that the cluster doesn't suffer any interruption.

> your database will be stopped then restarted on a new host
That's simply not true.

> During the operation the connection is dropped between your app and the database
It depends (cf below).

This page should be updated the same way as it was for PG and MySQL, eg.

Image

sc-zenokerr:
Fixes #3283

Updates the downtime information for redis during plan change.

Question: Do we need to update mongodb / elasticsearch and influxdb information too?

### DIFF
--- a/src/_posts/databases/redis/2000-01-01-start.md
+++ b/src/_posts/databases/redis/2000-01-01-start.md
@@ -1,7 +1,7 @@
 ---
 title: Scalingo for Caching
 nav: Introduction
-modified_at: 2025-06-06 00:00:00
+modified_at: 2025-12-29 00:00:00
 tags: databases redis addon
 index: 1
 ---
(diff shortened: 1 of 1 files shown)


Frzk (approved):
LGTM

PR #3476, opened 2025-12-30
fix(redis-haproxy): add warning about HAProxy connection limit

### LINKED ISSUE #3391: Document our HAProxy connection limit in DB HA cluster
Our HAProxy used for Redis HA cluster (and probably other technologies too) are limited to 2,000 TCP connections by default. Not a lot of customers would reach this limit, but once the limit is reached, TCP connections are abruptly refused (which is quite hard to investigate):
```
Redis::CannotConnectError : Connection timed out
```

HAProxy doesn't produce any log with our setup when such connections are refused.

sc-zenokerr:
Fixes #3391

Question: I am not sure this is the right part of the document to add this. Where do you think is best?

image

### DIFF
--- a/src/_posts/databases/redis/2000-01-01-start.md
+++ b/src/_posts/databases/redis/2000-01-01-start.md
@@ -439,7 +439,6 @@ If your database has multiple nodes, the dump is done on the secondary node.
 
 ### HAProxy Connection Limitation
 
-{% warning %}
 When using a Business plan, the HAProxy instances serving as entry points to your cluster have a default limit of **2,000 concurrent TCP connections**.
 
 If this limit is reached, any new connection will be abruptly refused.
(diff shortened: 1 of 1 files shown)


src/_posts/databases/redis/2000-01-01-start.md
```diff
@@ -435,4 +435,18 @@ If your database has multiple nodes, the dump is done on the secondary node.
...
+
+{% warning %}
+When using a Business plan, the HAProxy instances serving as entry points to your cluster have a default limit of **2,000 concurrent TCP connections**.
+
+If this limit is reached, any new connection will be abruptly refused.
+
+This can result in "Connection timed out" errors such as:
+
+- _Redis::CannotConnectError : Connection timed out_
+{% endwarning %}
```
Frzk: I'd remove the warning block:

```suggestion
When using a Business plan, the HAProxy instances serving as entry points to your cluster have a default limit of **2,000 concurrent TCP connections**.

If this limit is reached, any new connection will be abruptly refused. This can result in errors such as:

```shell
Redis::CannotConnectError : Connection timed out
```

```

sc-zenokerr: Making that change results in the notice being mixed with the redis trademark notice:
image

I could add a trademark notice heading (which appears in the menu and search etc) like this:
image

Or something else. WDYT?

sc-zenokerr: BTW. I haven't pushed any changes... I'm looking for your opinion first!

Frzk: @benjaminach: I'm wondering if the notice is still mandatory... Any thoughts about it?

benjaminach: I think it's preferable to keep this mention.

PR #3622, opened 2026-03-23
Add a tutorial to deploy Keycloak in cluster mode

Frzk:


EtienneM (changes requested):
praise: that's great, I don't have much to say about this :)

src/_tutorials/keycloak/index.md
```diff
@@ -0,0 +1,743 @@
...
+  ip_hash;
+}
+
+server {
+  server_name localhost;
+  listen <%= ENV["PORT"] %>;
+  charset utf-8;
+
+  # Optional hardening:
+  proxy_hide_header X-Powered-By;
```
yanjost, line 330: issue (high — security): this nginx sample doesn't set any `X-Forwarded-*` headers, but step 5 above mandates `KC_PROXY_HEADERS=xforwarded`, which tells Keycloak to trust the incoming `X-Forwarded-For` verbatim for client-IP resolution.

nginx's default is to forward whatever `X-Forwarded-For` the downstream sent — so a remote client can just set `X-Forwarded-For: 8.8.8.8` (rotating values too) and Keycloak will attribute the request to that fake IP. This defeats:

- the per-IP brute-force / `loginFailures` counter;
- realm-level IP allow/deny policies;
- audit-log forensics on `clientAddress`;
- any downstream SIEM rule keyed on Keycloak's reported source IP.

Suggested additions in the `server { … }` block (around the `# Optional hardening:` lines):

```suggestion
  # Optional hardening:
  proxy_hide_header X-Powered-By;

  # Trust only the Scalingo router for real-IP / XFF rewriting.
  # Replace with the documented Scalingo router CIDR(s).
  # set_real_ip_from <scalingo-router-CIDR>;
  # real_ip_header X-Forwarded-For;
  # real_ip_recursive on;

  proxy_set_header Host              $host;
  proxy_set_header X-Real-IP         $remote_addr;
  proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
  proxy_set_header X-Forwarded-Proto $scheme;
  proxy_set_header X-Forwarded-Host  $host;
  proxy_set_header X-Forwarded-Port  $server_port;
```

To verify afterwards: send a request with a forged `X-Forwarded-For` header, then check `clientAddress` in the Keycloak admin event log — it should show your real source IP, not the forged one.

yanjost (commented):
One security flag on the nginx sample — happy to chat about the exact Scalingo router CIDR if useful.

src/_tutorials/keycloak/index.md
Frzk: questions:
- Isn't `X-Forwarded-For` set by our frontend reverse proxies?
- If this is a security flaw, shouldn't we better put these settings in our default nginx config file (the one provided by our buildpack)?
- Can you provide the Scalingo routers CIDRs?

Thanks!

yanjost: Good questions — I traced the full request path, and I need to correct part of my initial comment.

Our front routers do set `X-Forwarded-For`, but the versioned router configuration preserves any client-provided value and appends the address obtained from the PROXY protocol. The resulting chain can therefore be `fake, real-client-IP`.

The nginx-buildpack correctly turns the last XFF element into `$remote_addr`. However, the tutorial’s nginx snippet then forwards the original XFF unchanged. Keycloak 26.5.1, through Quarkus 3.27.1, uses the first XFF element, so it would retain `fake`.

My original `$proxy_add_x_forwarded_for` suggestion was therefore incorrect: it would produce `fake, real, real`. The proxy must overwrite the headers:

```nginx
proxy_set_header Host              $host;
proxy_set_header X-Real-IP         $remote_addr;
proxy_set_header X-Forwarded-For   $remote_addr;
proxy_set_header X-Forwarded-Host  $host;
proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
proxy_set_header X-Forwarded-Port  $http_x_forwarded_port;
```

I think the immediate fix belongs in this tutorial because this specific Keycloak + application-proxy composition creates the unsafe trust chain. A broader nginx-buildpack hardening review may still be worthwhile, but it has compatibility implications for custom and multi-proxy configurations.

One more correction to my initial impact assessment: Keycloak’s native brute-force counter is user-based, not IP-based. The demonstrated impact is on audit/forensics and any IP-based policy, extension or SIEM processing.

src/_tutorials/keycloak/index.md
```diff
@@ -0,0 +1,628 @@
...
+4. Redeploy to Scalingo.
+
+### Environment
+
+Keycloak supports [many environment variables][kc-config].
+
+Moreover, the buildpack makes use of the following environment variables. They
+can be leveraged to customize your deployment:
+
+- `KEYCLOAK_VERSION`\\
```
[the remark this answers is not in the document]
EtienneM: Done

EtienneM (approved):
Nice! 🤩

src/_tutorials/keycloak/index.md
yanjost: I apologize for the comments being too verbose, as I'm not an expert in those topics, I did an AI-assisted analysis to be sure I correctly followed the path between Scalingo router -> Nginx -> Keycloak

Thanks for applying the fix, LGTM now!

yanjost (approved):
LGTM

PR #3666, opened 2026-04-09
Update analytics endpoint to new domain and path

leo-scalingo:
Related to http://github.com/Scalingo/api/issue/3232


aurelien-reeves-scalingo (approved):
If `assets/analytics.js` is still good, LGTM 👍

PR #3668, opened 2026-04-09
feat(menu): don't auto open submenus anymore

Frzk:


benjaminach (approved):
It makes navigation in a crowded service a bit easier.

PR #3679, opened 2026-04-16
docs(changelog): CLI v1.44.1

EtienneM:


sc-david-voisin (commented):
suggestion: 
not properly the subject, but what about to remove the superfluous line in v1.44.0 ?
"build(deps): update github.com/Scalingo/go-scalingo from v10 to v11"

otherwise LGTM

EtienneM (commented):
Good idea :)

sc-david-voisin (approved):

PR #3745, opened 2026-05-11
fix(postgresql/guides) plan change process is restarting follower instances first

curzolapierre:


src/_posts/databases/postgresql/shared-resources/guides/2000-01-01-changing-plan.md
```diff
@@ -45,8 +45,8 @@ though.
 
 ### From Starter to Business
 
-When changing for a Business plan, the platform first starts a follower
-instance with the targeted size. If necessary, it then reboots the primary
+When changing for a Business plan, the platform starts additional instances
+with the targeted size. If necessary, it then reboots the previously existing
```
Frzk: question:
When changing from a Starter plan to a Business one, we switch from a single-instance setup to a primary+follower setup, don't we?
So "additional instances" is the new "follower instance", and "previously existing instance" it the "primary" one, aren't they?
(if so, I don't get why this change)

src/_posts/databases/postgresql/shared-resources/guides/2000-01-01-changing-plan.md
```diff
@@ -71,7 +71,7 @@ though.
 ### From Business to Business
 
 When changing the size of a Business plan, the platform first reboots the
-primary instance with the targeted new size. It then reboots the follower
+follower instances with the targeted new size. It then reboots the primary
```
Frzk: question: is this new (or were we wrong 😨)?

curzolapierre: Yes this is new as mentioned here https://github.com/Scalingo/documentation/pull/3745/#ref-issue-4314886585

src/_posts/databases/postgresql/shared-resources/guides/2000-01-01-changing-plan.md
curzolapierre: > When changing from a Starter plan to a Business one, we switch from a single-instance setup to a primary+follower setup, don't we?

Yes correct

>  So "additional instances" is the new "follower instance", and "previously existing instance" it the "primary" one, aren't they?

To be more precise: "additional instances" are the new "follower instance(s)". This is the only little change that I wanted to bring. But from my first read the change was obvious, now it's not the case anymore 😅 

> it the "primary" one, aren't they?

Yes correct


So I'm suggesting a change: a24fdecf, WDYT?

Frzk (approved):
LGTM

PR #3762, opened 2026-05-27
Update Elixir buildpack URL in documentation

Fassih21:
This PR updates the Elixir buildpack URL in the documentation to the currently maintained Scalingo-compatible version.

The previous link was outdated and has been replaced with the correct community buildpack URL used in the sample application.

No other changes were made.


benjaminach:
Hey, thanks for your contribution.

We decided to remove Elixir from the documentation because support currently relies on a community-maintained buildpack, which may support newer scalingo-stack versions.

The buildpack should continue to work, but since we don’t maintain or validate it ourselves, we can’t officially document or support it at this time.

That being said, adapting the buildpack maintained by Gigalixir to make it compatible with our naming conventions would also work: https://github.com/gigalixir/gigalixir-buildpack-elixir

PR #3796, opened 2026-06-09
Generalize inline code padding inside and outside notes

benjaminach:
# Before
Screenshot 2026-06-09 at 18 00 42

# After
Screenshot 2026-06-09 at 18 00 36


baptmaury (approved):
LGTM, nice

Frzk (approved):
Thanks 🤩

PR #3799, opened 2026-06-10
fix: rename properly changelog posts filenames

sc-david-voisin:
Fix latest Ubuntu stack changes order within changelog posts.

related to https://scalingo.slack.com/archives/C01PFD3GBU7/p1781098269634679


benjaminach:
If we release scalingo-26 today, I will update the date to 10 June.

benjaminach (changes requested):
I would modify the publication date of the scalingo-26 changelog note to today.

sc-brandon:
> I would modify the publication date of the scalingo-26 changelog note to today.

It is already released. Why modifying the date?

sc-brandon (approved):

sc-david-voisin:
> > I would modify the publication date of the scalingo-26 changelog note to today.
> 
> It is already released. Why modifying the date?

Ah, I've just pushed the modification (easy to revert).
Do we have a source of truth somewhere ? Or maybe we just want to announce it today as well ?

benjaminach (approved):
The date in the filename should reflect when we want customers to be notified about this change. If it has been available since 28 May, LGTM.

sc-david-voisin:
For clarity, we agreed with @benjaminach  to keep 10th of June, to maximize the visibility (even if it was released before).

PR #3848, opened 2026-07-03
fix(changelog): add breadcrumb labels and use canonical detail links 

benjaminach:
## Summary

Fix changelog detail page breadcrumbs and expose canonical changelog detail URLs in the changelog HTML.

## Details

Changelog entries live outside `src/_posts`, so their parent paths were not registered in `breadcrumb_hash`. As a result, changelog detail pages rendered an empty breadcrumb like ` - ` above the title.

This PR registers only the missing changelog breadcrumb labels:

- `/changelog` → `Changelog`
- `/changelog/<category>` → formatted category name

Changelog categories are rendered as plain text in breadcrumbs because category URLs like `/changelog/databases` do not exist and should not be clickable.

This PR also updates the changelog list checkbox `id` and matching `label for` to use `post.url` instead of `post.id | url_decode`, so the HTML exposes the canonical Jekyll URL with hyphens, for example:

- `/changelog/api/continuous-backup-events`

instead of:

- `/changelog/api/continuous_backup_events`

### DIFF
--- a/src/_plugins/jekyll-breadcrumb-tag.rb
+++ b/src/_plugins/jekyll-breadcrumb-tag.rb
@@ -13,7 +13,11 @@ module Jekyll
         new_path += "/#{part}"
         title = breadcrumb_hash[new_path]
 
-        "<a href=\"#{new_path}\">#{title}</a>"
+        if new_path.start_with?("/changelog/")
+          title
+        else
+          "<a href=\"#{new_path}\">#{title}</a>"
+        end
       end.join(" - ")
     end
   end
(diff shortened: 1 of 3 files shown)


Frzk (approved):
LGTM

EtienneM (changes requested):
I'm sorry, I fail to understand which breadcrumb this PR is about. Could you add screenshots to highlights your changes?

benjaminach:
When accessing a changelog not detail page, the breadcrumb area is empty.

Screenshot 2026-07-22 at 09 26 14

After the change, the breadcrumb is populated with the correct path
Screenshot 2026-07-22 at 09 27 11

EtienneM (approved):

PR #3857, opened 2026-07-09
docs(backup-policies): explain there may be a delay between the configured time and the actual time

### LINKED ISSUE #3856: [Backup Policies] Make it more explicit that the configured time is best effort
https://doc.scalingo.com/databases/about/backup-policies#scheduled-backups

We should explicitly write that even though the time of the backup can be configured, the actual time of the backup may be later.

EtienneM:
Fix #3856


Frzk (approved):
LGTM

PR #3876, opened 2026-07-27
Clarify application swap documentation

benjaminach:


benjaminach:
@mickael-plassard do you think these updated formulation may be clearer?

EtienneM (approved):

PR #3879, opened 2026-07-30
Improve the pulling of repository for tutorials and demo

SC-Samir:
This PR permit to only pull the folder needed for the tutorial using git sparse-checkout.

### DIFF
--- a/src/_tutorials/pgvector-recommendation/index.md
+++ b/src/_tutorials/pgvector-recommendation/index.md
@@ -45,8 +45,11 @@ The recommendation workflow follows these steps:
 1. Clone our repository:
 
    ```bash
-   git clone https://github.com/Scalingo/scalingo-labs
-   cd scalingo-labs/filmrecommendation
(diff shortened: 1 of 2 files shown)


src/_tutorials/pgvector-recommendation/index.md
EtienneM, line 1: question: is it really interesting to do that? I feel like the new version is harder to understand, don't you think?

EtienneM (changes requested):

src/_tutorials/pgvector-recommendation/index.md
SC-Samir: It was a @Frzk suggestion, to avoid pulling the full github repository

EtienneM (approved):
I'm not keen on this choice but I let you the final decision.

Frzk:
I found a bit unfortunate to clone the whole repo for a simple demo.
But I've no strong opinion about this. Maybe it's better to keep instructions simple :)

PR #3963, opened 2026-09-11
[INC-881] Do not specify .git for multi-buildpack example keeps default behavioour (downloading tarballs)

leo-scalingo:


Frzk (approved):
LGTM
question: shall we prepare a changelog entry or broader communication as well?

leo-scalingo:
@Frzk we'll do if the problem comes too often (we could even automate the fix) For now, no more work :)

PR #3965, opened 2026-09-14
feat: Try new configuration of crawler and index

SC-Samir:
This commit permit to implement the solution given by algolia support to solve our search problem


assets/application.js
```diff
@@ -27,7 +27,7 @@ const initialQuery = urlParams.get('q');
       container: container,
       appId: 'RWJM2H1BD2',
       apiKey: '9dfb74cc002ece507fac441e93da6345',
-      indexName: 'scalingo-doc',
+      indexName: 'scalingo_test',
```
benjaminach: Why "test" 😭 Couldn't we have something better like 'scalingo\_index" or '_prod'?

SC-Samir: this PR is only for testing the new search configuration on the review app, when the solution will be validated. I will change the name of the index

SC-Samir: done

Frzk (approved):

PR #3969, opened 2026-09-14
fix(tutorials): fix page layout

Frzk:
Rely on gap instead of margins to keep vertical spacing homogeneous:
image


leo-scalingo (approved):
LGTM

PR #3972, opened 2026-09-15
docs(sdk+cli): Document go-scalingo 11.4.0 and CLI 1.49.0

matthieu526-scalingo:


EtienneM (approved):
Why "tasks" in the PR description? It's not in our list of keywords and I don't see how it relates to the content of this PR. "docs" is probably more accurate

matthieu526-scalingo:
> Why "tasks" in the PR description? It's not in our list of keywords and I don't see how it relates to the content of this PR. "docs" is probably more accurate

Because I'm not sure which one to use, especially just to publish a release. But I'll replace with `docs`.

PR #3976, opened 2026-09-17
docs(postgresql): clarify Dedicated Resources availability

benjaminach:
## Summary
- notify readers on the PostgreSQL Dedicated Resources provisioning page that this architecture model is currently limited to selected customers
- reuse the same availability notice across related Dedicated Resources documentation pages, so the message stays consistent


mcimbaro (approved):
LGTM

